PR #38, opened 2021-12-08
Add example of AKS attestation and secret provisioning

veenasai2:
Signed-off-by: Veena Saini <veena.saini@intel.com>

## Description of the changes 

This PR provides a reference implementation to show how gramine attestation (DCAP) samples works inside AKS cluster. We have created two docker images for ra-tls-secret-prov server and ra-tls-secret-prov client. Both images are deployed as part of AKS confidential compute cluster and both quote generation and quote verification are successful inside AKS cluster.

For client deployment inside AKS cluster, we have gsc/Examples/aks-attestation/aks-secret-prov-client-deployment.yaml and for server  deployment gsc/Examples/aks-attestation/aks-secret-prov-server-deployment.yaml file.

For more details, we have created a readme file.  

This PR is an updated version of https://github.com/gramineproject/gsc/pull/11.

## How to test this PR? 
Please follow gsc/Examples/aks-attestation/README.md


dimakuv (changes requested):
> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @veenasai2)

___
*[-- commits, line 3 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU3dg6-PtsqmJUKn70):*
I'd like to add more explanations in this commit message. Taking from PR, smth like this:
```
This commit provides a reference implementation to show how Gramine DCAP attestation examples work in an Azure Kubernetes Service (AKS) cluster. The added example creates two Docker images: one for (non-SGX) `ra-tls-secret-prov` server and one for SGX `ra-tls-secret-prov` client. Both images are deployed as part of the AKS Confidential Compute cluster.
```

We'll do it during final rebase.
___
*[Examples/aks-attestation/README.md, line 18 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU566kCpNwykBVb09C:-MqU566kCpNwykBVb09D:b-chn1kh) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L18)):*
> ```Markdown
> server applications, where by default server is running on localhost:4433. In the example, the
> client sends its SGX quote to the server for verification. After successful quote verification, the
> server sends a secret to the client. To run these client and server applications inside the AKS
> ```

Please start `To run these client ...` sentence as a new paragraph.
___
*[Examples/aks-attestation/README.md, line 19 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU5BmlAUfYHi9Yoc5l:-MqU5BmlAUfYHi9Yoc5m:b-g1tvn0) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L19)):*
> ```Markdown
> client sends its SGX quote to the server for verification. After successful quote verification, the
> server sends a secret to the client. To run these client and server applications inside the AKS
> cluster, user needs to prepare two docker images, one for the client and one for the server. In our
> ```

`docker` -> `Docker`
___
*[Examples/aks-attestation/README.md, line 28 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU5f6rAX0mFR3IKkIc:-MqU5f6rAX0mFR3IKkId:bwx9i7g) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L28)):*
> ```Markdown
> `localhost` to `<AKS-DNS-NAME.*.cloudapp.azure.com>`.
> 
> In order to create base client and server images for the AKS environment, user can execute the
> ```

`images` -> `Docker images`
___
*[Examples/aks-attestation/README.md, line 61 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU5tLw-cSmb685cDW2:-MqU5tLw-cSmb685cDW3:b-u2liji) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L61)):*
> ```Markdown
> 2. Create the GSC client image:
> 
>     Note: We tested this example with DCAP driver 1.11 specified in the GSC configuration file.
> ```

This looks a bit ugly. Let's re-format like this:
```
2. Create the GSC client image (note that we tested this example with DCAP driver 1.11 specified in the GSC configuration file):
```

Please re-wrap my proposed text to 80-char limit.
___
*[Examples/aks-attestation/README.md, line 66 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU6CpN2NWTvqC0GaUN:-MqU6CpN2NWTvqC0GaUO:b-qhrfi7) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L66)):*
> ```Markdown
>     $ cd gsc
>     $ cp config.yaml.template config.yaml
>     $ openssl genrsa -3 -out enclave-key.pem 3072
> ```

Actually, let's remove these two commands (cp and openssl genrsa). The GSC user is supposed to have done this already. No need to replicate it in this example.
___
*[Examples/aks-attestation/README.md, line 95 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU6qP37FXBNSPjjOAZ:-MqU6qP37FXBNSPjjOA_:b-ha42zy) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L95)):*
> ```Markdown
> service for the container node. The service will internally connect with az-dcap-client to fetch the
> platform collateral required for quote generation. In this demo, the
> `aks-secret-prov-client-deployment.yaml` uses aesmd service exposed by AKS with the help of the
> ```

```
`aks-secret-prov-client-deployment.yaml` uses
```
->
```
`aks-secret-prov-client-deployment.yaml` file uses
```
___
*[Examples/aks-attestation/README.md, line 96 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU6hcA9_6pmle2v3hO:-MqU6hcA9_6pmle2v3hP:b-xnocq6) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L96)):*
> ```Markdown
> platform collateral required for quote generation. In this demo, the
> `aks-secret-prov-client-deployment.yaml` uses aesmd service exposed by AKS with the help of the
> sgxquotehelper plugin.
> ```

We have `AESMD` in the very first paragraph of this README. Let's be consistent and also replace `aesmd` -> `AESMD` everywhere in this paragraph.

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: 12 of 13 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[-- commits, line 3 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU3dg6-PtsqmJUKn70):*

I'd like to add more explanations in this commit message. Taking from PR, smth like this:
```
This commit provides a reference implementation to show how Gramine DCAP attestation examples work in an Azure Kubernetes Service (AKS) cluster. The added example creates two Docker images: one for (non-SGX) `ra-tls-secret-prov` server and one for SGX `ra-tls-secret-prov` client. Both images are deployed as part of the AKS Confidential Compute cluster.
```

We'll do it during final rebase.
</blockquote></details>

ok, thanks.
___
*[Examples/aks-attestation/README.md, line 18 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU566kCpNwykBVb09C:-MqYOYWJ-RMxyTzIWF3a:b-lhs69o) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L18)):*

Please start `To run these client ...` sentence as a new paragraph.
</blockquote></details>

done, thanks
___
*[Examples/aks-attestation/README.md, line 19 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU5BmlAUfYHi9Yoc5l:-MqYOd550cC9sx8uHFe-:b-lhs69o) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L19)):*

`docker` -> `Docker`
</blockquote></details>

done, thanks
___
*[Examples/aks-attestation/README.md, line 28 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU5f6rAX0mFR3IKkIc:-MqYOlH144JYE8Jb-WdE:b-lhs69o) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L28)):*

`images` -> `Docker images`
</blockquote></details>

done, thanks
___
*[Examples/aks-attestation/README.md, line 61 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU5tLw-cSmb685cDW2:-MqYP9ll25C7VkznMFt_:b-m6nt8b) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L61)):*

This looks a bit ugly. Let's re-format like this:
```
2. Create the GSC client image (note that we tested this example with DCAP driver 1.11 specified in the GSC configuration file):
```

Please re-wrap my proposed text to 80-char limit.
</blockquote></details>

80 or 100 char limit?
___
*[Examples/aks-attestation/README.md, line 66 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU6CpN2NWTvqC0GaUN:-MqYPI648NHIDTOKYXFG:b-wpb374) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L66)):*

Actually, let's remove these two commands (cp and openssl genrsa). The GSC user is supposed to have done this already. No need to replicate it in this example.
</blockquote></details>

ok, thanks.
___
*[Examples/aks-attestation/README.md, line 95 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU6qP37FXBNSPjjOAZ:-MqYPaMv8VbEzxjUiqTE:b-lhs69o) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L95)):*

```
`aks-secret-prov-client-deployment.yaml` uses
```
->
```
`aks-secret-prov-client-deployment.yaml` file uses
```
</blockquote></details>

done, thanks
___
*[Examples/aks-attestation/README.md, line 96 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU6hcA9_6pmle2v3hO:-MqYPeqA6_iMaSjZXvHw:b-r18wie) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L96)):*

We have `AESMD` in the very first paragraph of this README. Let's be consistent and also replace `aesmd` -> `AESMD` everywhere in this paragraph.
</blockquote></details>

done, thanks.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @veenasai2)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/38#-MqmlOsY7q8JcsCn2Pt8:-MqmlOsY7q8JcsCn2Pt9:b5avb0w) (no related file):*
Bar the commit message change, I'm fine with this PR. Approving.

---
___
*[Examples/aks-attestation/README.md, line 61 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU5tLw-cSmb685cDW2:-Mqmkh3QDtw0dnw54LjW:ba0sseb) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L61)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

80 or 100 char limit?
</blockquote></details>

Sorry, you have 100 char limit in this document, so you should have aligned to 100 char limit. Which you did. Anyway, resolved.

(In our RST documentation, we use 80 char limit. However, for MD (Markdown) documentation of our examples, we don't impose a strict limit, so using 100 char limit is fine.)

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/38#-MqmlOsY7q8JcsCn2Pt8:-MqmlbvVCGXzm9X-vIKQ:b-wpb374) (no related file):*

Bar the commit message change, I'm fine with this PR. Approving.
</blockquote></details>

ok, thanks.

---
___
*[Examples/aks-attestation/README.md, line 61 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MqU5tLw-cSmb685cDW2:-Mqml_Uq1XtBYNO60qbL:b-wpb374) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/README.md#L61)):*

Sorry, you have 100 char limit in this document, so you should have aligned to 100 char limit. Which you did. Anyway, resolved.

(In our RST documentation, we use 80 char limit. However, for MD (Markdown) documentation of our examples, we don't impose a strict limit, so using 100 char limit is fine.)
</blockquote></details>

ok, thanks.

aneessahib:
@mkow @woju pls review. We had beaten this up royally over many weeks so hopefully should be a quick one for you :)

mkow (changes requested):
> Reviewed 12 of 13 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: all files reviewed, 19 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @veenasai2)

___
*[Examples/aks-attestation/aks-secret-prov-client-deployment.yaml, line 13 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msjql-K54vbeZvW2bM5:-Msjql-K54vbeZvW2bM6:b-v5ylcw) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-client-deployment.yaml#L13)):*
> ```YAML
>            app: gsc-ra-tls-secret-prov-client
>     spec:
>       volumes:
> ```

```
spec:
  volumes:
```

```
metadata:
   labels:
```

```
labels:
    app: gsc-ra-tls-secret-prov-client
```

Please decide on indentation and unify it (here and in the other YAMLs).
___
*[Examples/aks-attestation/aks-secret-prov-client-deployment.yaml, line 22 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msjs88J1ntC6eCymPUZ:-Msjs88J1ntC6eCymPU_:b1bhpx4) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-client-deployment.yaml#L22)):*
> ```
> ra-tls-server-aks-dns.eastus.cloudapp.azure.com:4433
> ```

What's this domain?
___
*[Examples/aks-attestation/aks-secret-prov-server.dockerfile, line 16 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msk8quS3H9N3WXBNklL:-Msk8quTDSj_C6FNDPdi:bbwakqu) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-server.dockerfile#L16)):*
> ```Dockerfile
> # version used for quote generation. User can replace the below package with the
> # latest package.
> RUN wget https://packages.microsoft.com/ubuntu/18.04/prod/pool/main/a/az-dcap-client/az-dcap-client_1.10_amd64.deb \
> ```

Any reason for installing this particular deb manually, but others via `apt`?
___
*[Examples/aks-attestation/aks-secret-prov-server-deployment.yaml, line 25 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msk9M-BDgF7Ubo6UmMK:-Msk9M-BDgF7Ubo6UmML:bm0wn7g) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-server-deployment.yaml#L25)):*
> ```YAML
> 
> # TODO: Requesting an SGX machine is not needed, but Intel DCAP libraries have a bug of trying to
> # open the SGX driver (see  https://github.com/intel/linux-sgx/issues/756)
> ```

double space
___
*[Examples/aks-attestation/aks-secret-prov-server-deployment.yaml, line 25 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msk9Qjc2M0Cw0GJVgi4:-Msk9Qjd22FSf-0jRi9t:b-jkw77d) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-server-deployment.yaml#L25)):*
> ```YAML
> 
> # TODO: Requesting an SGX machine is not needed, but Intel DCAP libraries have a bug of trying to
> # open the SGX driver (see  https://github.com/intel/linux-sgx/issues/756)
> ```

Seems this bug is already fixed?
___
*[Examples/aks-attestation/base-image-generation-script.sh, line 6 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskCoRODvDlv5A_8PW9:-MskCoRODvDlv5A_8PWA:b-bdvelz) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/base-image-generation-script.sh#L6)):*
> ```Shell
>     echo "***** gramine directory exists, proceeding to image generation *****"
> else
>     bash ./gramine_build.sh
> ```

Hmm, how does this work? Do we build Gramine twice and separately, once here and once in the GSC code? Why? What if both Gramine versions differ?
___
*[Examples/aks-attestation/gramine_build.sh, line 2 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskA-f9AM_PNkmi0Omc:-MskA-f9AM_PNkmi0Omd:b62tvf9) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/gramine_build.sh#L2)):*
> https://gramine.readthedocs.io/en/latest/building.html#id2

dead link
___
*[Examples/aks-attestation/gramine_build.sh, line 4 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskA5viA9ibpQmfmopU:-MskA5viA9ibpQmfmopV:b-jlqd7w) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/gramine_build.sh#L4)):*
> ```Shell
> # Please refer to https://gramine.readthedocs.io/en/latest/building.html#id2 for more details.
> 
> # install Gramine dependencies
> ```

Other comments here are capitalized, please unify
___
*[Examples/aks-attestation/gramine_build.sh, line 27 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskAMs4BspriTELiHtA:-MskAMs4BspriTELiHtB:bea1hmd) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/gramine_build.sh#L27)):*
> ```Shell
> 
> # Download Gramine
> git clone https://github.com/gramineproject/gramine.git
> ```

Please add `--depth=1`.
___
*[Examples/aks-attestation/gramine_build.sh, line 31 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskAb2T7saRou8Ktyge:-MskAb2T7saRou8Ktygf:b-ncl30q) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/gramine_build.sh#L31)):*
> ```Shell
> mkdir -p meson_build_output
> 
> # Generate Signing Key
> ```

`Signing Key` -> `signing key`
___
*[Examples/aks-attestation/README.md, line 1 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskDrzk3Pc33Ffu6FCG:-MskDrzk3Pc33Ffu6FCH:b66o8dz) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L1)):*
> Gramine Attestation Inside AKS cluster

Here and in all other places: You seem to capitalize words quite randomly ;) Please change to capitalizing only the first word and proper names.
___
*[Examples/aks-attestation/README.md, line 28 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskB1rf7MbgnQSJOoxY:-MskB1rf7MbgnQSJOoxZ:bi0zpdx) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L28)):*
> `<AKS-DNS-NAME.*.cloudapp.azure.com>`

I'm a bit uneasy about this wildcard. Shouldn't we bind to a specific domain, or a list of them instead? Is it possible to create an instance on Azure which has a colliding <AKS-DNS-NAME>, but a different wildcarded part?
___
*[Examples/aks-attestation/README.md, line 32 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskGHW80BCI3NlmipDO:-MskGHW80BCI3NlmipDP:b-2i9drb) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L32)):*
> Since both client and server applications will run inside containers in the AKS cluster

What's the point of using SGX then? :thinking:
If you want to remove the need for trusting the cloud provider then you can't run the verifier on the very same infrastructure?
___
*[Examples/aks-attestation/README.md, line 54 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskHH020iGfz0-BZXUH:-MskHH020iGfz0-BZXUI:b7cduiv) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L54)):*
> ```
>     - Reference deployment file:
>         `aks-secret-prov-server-deployment.yaml`
> ```

Is this a code snippet for some AKS configuration file? If so, it should be wrapped in triple backticks. I'm not really sure how to interpret it in the current form, which is a single-entry bullet list.
___
*[Examples/aks-attestation/README.md, line 81 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskHnIkBMVUQJZ7olRX:-MskHnIl7yHvi3nik2vF:b62r0uv) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L81)):*
> ```Markdown
> 6. Deploy `gsc-aks-secret-prov-client-img:latest` in AKS confidential compute cluster:
>     - Reference deployment file:
>         `aks-secret-prov-client-deployment.yaml`
> ```

ditto (formatting)
___
*[Examples/aks-attestation/README.md, line 124 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskJ8uSDgvRc2-X490d:-MskJ8uSDgvRc2-X490e:b-12z95i) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L124)):*
> ```
> the secret has been provisioned to the client
> ```

Wait, who's the client and who's the server here? Usually it's the server which runs inside SGX providing some service and the clients can connect to it and attest it. Here the clients are SGX workers which receive jobs from a trusted server after verification? Or is it some other model?
___
*[Examples/aks-attestation/certs/README, line 8 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskKIss9XeIMNZEJxDz:-MskKIss9XeIMNZEJxE-:brlonrc) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/certs/README#L8)):*
> ```
>   Loaded in server (verifier), so it will send it to the client during TLS
>   handshake. The "Common Name" field is set to
>   `ra-tls-server-aks-dns.eastus.cloudapp.azure.com`.
> ```

Why include such a certificate in the repo? It's unusable by anyone except the author of this PR, which controls `ra-tls-server-aks-dns.eastus.cloudapp.azure.com` (as I assume). Or maybe I'm missing something?
___
*[Examples/aks-attestation/certs/README, line 9 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskKXL7DjaYGjhEVoTU:-MskKXL7DjaYGjhEVoTV:bcerdm0) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/certs/README#L9)):*
> ```
>   handshake. The "Common Name" field is set to
>   `ra-tls-server-aks-dns.eastus.cloudapp.azure.com`.
> - `server2.key` -- RSA private key in PEM format. Loaded in server (verifier).
> ```

This should be generated during build / provided as argument instead of including in the repo, otherwise people may accidentally deploy it somewhere.

veenasai2 (commented):
Thank you very much @mkow for reviewing this PR. For now I am just answering your queries. I will soon push all the suggested changes. Thanks.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38#-:-MsnpeTF2xeHn2ZKWSoG:bv5kgei)* status: all files reviewed, 19 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Examples/aks-attestation/aks-secret-prov-client-deployment.yaml, line 13 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msjql-K54vbeZvW2bM5:-MsnetGHD2UYTOTvwkmI:b77i4af) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-client-deployment.yaml#L13)):*

```
spec:
  volumes:
```

```
metadata:
   labels:
```

```
labels:
    app: gsc-ra-tls-secret-prov-client
```

Please decide on indentation and unify it (here and in the other YAMLs).
</blockquote></details>

sure, thanks.
___
*[Examples/aks-attestation/aks-secret-prov-client-deployment.yaml, line 22 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msjs88J1ntC6eCymPUZ:-MsneNLs3D1V1Ih02tGv:bkg9xdf) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-client-deployment.yaml#L22)):*

> ```
> ra-tls-server-aks-dns.eastus.cloudapp.azure.com:4433
> ```

What's this domain?
</blockquote></details>

this domain name is assigned to the container where ra-tls-secret-prov server is running. We declare the domain name in aks-secret-prov-server-deployment.yaml file as  ``` service.beta.kubernetes.io/azure-dns-label-name: ra-tls-server-aks-dns``` 
___
*[Examples/aks-attestation/aks-secret-prov-server.dockerfile, line 16 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msk8quS3H9N3WXBNklL:-Msn7nvu3cuviIZlEnTg:b-kk1cuc) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-server.dockerfile#L16)):*

Any reason for installing this particular deb manually, but others via `apt`?
</blockquote></details>

In the past we noticed some regression with latest az-dcap-client versions. That's why we kept the one that is working for us. Thanks.
___
*[Examples/aks-attestation/aks-secret-prov-server-deployment.yaml, line 25 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msk9M-BDgF7Ubo6UmMK:-MsngoRs2RMlgWF5q5SS:bp9funo) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-server-deployment.yaml#L25)):*

double space
</blockquote></details>

sure, I will remove that.
___
*[Examples/aks-attestation/aks-secret-prov-server-deployment.yaml, line 25 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msk9Qjc2M0Cw0GJVgi4:-MsngrTI1mZSyVL0gyig:b-66irxt) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-server-deployment.yaml#L25)):*

Seems this bug is already fixed?
</blockquote></details>

Need to test this one again, thanks
___
*[Examples/aks-attestation/base-image-generation-script.sh, line 6 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskCoRODvDlv5A_8PW9:-MsngzEj1obuwc1xivRb:b-jks3jk) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/base-image-generation-script.sh#L6)):*

Hmm, how does this work? Do we build Gramine twice and separately, once here and once in the GSC code? Why? What if both Gramine versions differ?
</blockquote></details>

First step is to create ra-tls-secret-prov libraries which are later used for creating  base client and server images. Second step is where we graminize client image as part of gsc. Hence, being of different version will not impact each other. Thanks.
___
*[Examples/aks-attestation/gramine_build.sh, line 2 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskA-f9AM_PNkmi0Omc:-MsnDbZJ4ccL1rEjakag:bukhhgc) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/gramine_build.sh#L2)):*

> https://gramine.readthedocs.io/en/latest/building.html#id2

dead link
</blockquote></details>

Thanks I will replace with working link.
___
*[Examples/aks-attestation/gramine_build.sh, line 4 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskA5viA9ibpQmfmopU:-MsnDfLc9s1_K8blkJSJ:b77i4af) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/gramine_build.sh#L4)):*

Other comments here are capitalized, please unify
</blockquote></details>

sure, thanks.
___
*[Examples/aks-attestation/gramine_build.sh, line 27 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskAMs4BspriTELiHtA:-MsnFrAY9wF_DX8buLia:bwb726v) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/gramine_build.sh#L27)):*

Please add `--depth=1`.
</blockquote></details>

sure, thanks
___
*[Examples/aks-attestation/gramine_build.sh, line 31 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskAb2T7saRou8Ktyge:-MsnG0Os4ygmfPZNk_M7:bwb726v) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/gramine_build.sh#L31)):*

`Signing Key` -> `signing key`
</blockquote></details>

sure, thanks
___
*[Examples/aks-attestation/README.md, line 1 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskDrzk3Pc33Ffu6FCG:-MsnhzKD2jgsReLMh28g:b-wpb374) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L1)):*

> Gramine Attestation Inside AKS cluster

Here and in all other places: You seem to capitalize words quite randomly ;) Please change to capitalizing only the first word and proper names.
</blockquote></details>

ok, thanks.
___
*[Examples/aks-attestation/README.md, line 28 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskB1rf7MbgnQSJOoxY:-Msni3W6-XKwtURS3bAa:b-r7cdu2) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L28)):*

> `<AKS-DNS-NAME.*.cloudapp.azure.com>`

I'm a bit uneasy about this wildcard. Shouldn't we bind to a specific domain, or a list of them instead? Is it possible to create an instance on Azure which has a colliding <AKS-DNS-NAME>, but a different wildcarded part?
</blockquote></details>

This domain name is user configured domain name for his deployment. Depending upon the Azure subscription this domain could be anything (e.g. eastus, westus) , that's why we kept it open for the user to fill. User just need to provid the AKS-DNS-NAME in server deployment file and the complete dns name can be fetched from the azure portal post deployment.
___
*[Examples/aks-attestation/README.md, line 32 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskGHW80BCI3NlmipDO:-Msnj15UAmv0aC6IdDWl:b-hdxksu) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L32)):*

> Since both client and server applications will run inside containers in the AKS cluster

What's the point of using SGX then? :thinking:
If you want to remove the need for trusting the cloud provider then you can't run the verifier on the very same infrastructure?
</blockquote></details>

Although these are part of the same cluster but they may be located at geographically different machines. Hence the need for SGX attestation for a remote party.
___
*[Examples/aks-attestation/README.md, line 54 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskHH020iGfz0-BZXUH:-MsnkS3L9NIfmVhoXRSn:b90glvh) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L54)):*

> ```
>     - Reference deployment file:
>         `aks-secret-prov-server-deployment.yaml`
> ```

Is this a code snippet for some AKS configuration file? If so, it should be wrapped in triple backticks. I'm not really sure how to interpret it in the current form, which is a single-entry bullet list.
</blockquote></details>

This is not a code snippet. This is just a reference deployment file that the user can customise.
___
*[Examples/aks-attestation/README.md, line 81 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskHnIkBMVUQJZ7olRX:-Msnkbdl-SYowRMafcyn:by713e4) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L81)):*

ditto (formatting)
</blockquote></details>

This is not a code snippet. This is just a reference deployment file that user can customise.
___
*[Examples/aks-attestation/README.md, line 124 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskJ8uSDgvRc2-X490d:-MsnkjwH65-BU7kHxS8c:bfdxrs8) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L124)):*

> ```
> the secret has been provisioned to the client
> ```

Wait, who's the client and who's the server here? Usually it's the server which runs inside SGX providing some service and the clients can connect to it and attest it. Here the clients are SGX workers which receive jobs from a trusted server after verification? Or is it some other model?
</blockquote></details>

Step 1: Server (not necessarily running on SGX enabled machine): Waiting for connections. It has a domain name:port as  ra-tls-server-aks-dns.eastus.cloudapp.azure.com:4433

Step 2: Client (Running on SGX enabled machine): Sends its SGX quote to the above server.

After verifying the quote, the server will provision a dummy secret to the client. 

 
___
*[Examples/aks-attestation/certs/README, line 8 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskKIss9XeIMNZEJxDz:-Msnm-vE7-rBv2ureK6Q:bo7vj7r) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/certs/README#L8)):*

Why include such a certificate in the repo? It's unusable by anyone except the author of this PR, which controls `ra-tls-server-aks-dns.eastus.cloudapp.azure.com` (as I assume). Or maybe I'm missing something?
</blockquote></details>

This is just a sample certificate to get the demo working. We are explicitely calling out the fields the user would have to modify before generating their actual production certificates. 
___
*[Examples/aks-attestation/certs/README, line 9 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskKXL7DjaYGjhEVoTU:-Msno88EBxh0Zs9en73b:b-fjkp45) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/certs/README#L9)):*

This should be generated during build / provided as argument instead of including in the repo, otherwise people may accidentally deploy it somewhere.
</blockquote></details>

Actually here, I am following the same certficates directory format that we have for https://github.com/gramineproject/gramine/tree/master/CI-Examples/ra-tls-secret-prov/certs . Your point is valid. Here, we are providing the key just to reduce an extra step to the sample working. I can call out seperately in README that the key should not be used in production.

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: 9 of 13 files reviewed, 19 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Examples/aks-attestation/aks-secret-prov-server-deployment.yaml, line 25 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msk9M-BDgF7Ubo6UmMK:-Mt2O8A081wNeo7T7Fna:b-r18wie) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-server-deployment.yaml#L25)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

sure, I will remove that.
</blockquote></details>

done, thanks.
___
*[Examples/aks-attestation/aks-secret-prov-server-deployment.yaml, line 25 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msk9Qjc2M0Cw0GJVgi4:-Mt2OEg2DszXS2InIZ69:b-56vk8k) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-server-deployment.yaml#L25)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

Need to test this one again, thanks
</blockquote></details>

I am still able to see error logs for the server, if I don't use epc memory for the server container. As I can see from the issue (https://github.com/intel/linux-sgx/issues/756) comments that the fix will be merged in some upcoming releases and so far the latest release is on 18 Nov 2021. So , I am still keeping the TODO comment.  Thanks
___
*[Examples/aks-attestation/gramine_build.sh, line 2 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskA-f9AM_PNkmi0Omc:-Mt2O5LGAClA0_m9mEcA:b-r18wie) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/gramine_build.sh#L2)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

Thanks I will replace with working link.
</blockquote></details>

done, thanks.
___
*[Examples/aks-attestation/gramine_build.sh, line 4 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskA5viA9ibpQmfmopU:-Mt2O39h5h0WBEAjcNiY:b-r18wie) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/gramine_build.sh#L4)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

sure, thanks.
</blockquote></details>

done, thanks.
___
*[Examples/aks-attestation/gramine_build.sh, line 27 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskAMs4BspriTELiHtA:-Mt2O1VX7k2dq24GoLvJ:b-r18wie) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/gramine_build.sh#L27)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

sure, thanks
</blockquote></details>

done, thanks.
___
*[Examples/aks-attestation/gramine_build.sh, line 31 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskAb2T7saRou8Ktyge:-Mt2O-k6Eu-e4W8RszVu:b-lhs69o) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/gramine_build.sh#L31)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

sure, thanks
</blockquote></details>

done, thanks
___
*[Examples/aks-attestation/README.md, line 1 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskDrzk3Pc33Ffu6FCG:-Mt2MqIU5AtaaIzzohIU:b-r18wie) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L1)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

ok, thanks.
</blockquote></details>

done, thanks.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: 9 of 13 files reviewed, 19 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Examples/aks-attestation/README.md, line 32 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskGHW80BCI3NlmipDO:-MtE65S90e2Sli8wF17L:bta0o5v) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L32)):*
So, so the server attest the client (isn't this flipped? normally we call client the party that does the attestation)? Then who attests the server?

> Although these are part of the same cluster but they may be located at geographically different machines. Hence the need for SGX attestation for a remote party.

But you can just use TLS to have machine to machine verification, no need for SGX there.

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: 9 of 13 files reviewed, 19 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Examples/aks-attestation/README.md, line 32 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskGHW80BCI3NlmipDO:-MtGlWMf7sYhZ0nVU99x:b-4gq0s) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L32)):*
<details><summary><i>Previously, boryspoplawski (Borys Popławski) wrote…</i></summary><blockquote>

So, so the server attest the client (isn't this flipped? normally we call client the party that does the attestation)? Then who attests the server?

> Although these are part of the same cluster but they may be located at geographically different machines. Hence the need for SGX attestation for a remote party.

But you can just use TLS to have machine to machine verification, no need for SGX there.
</blockquote></details>

@boryspoplawski Here, the client wants to prove to the server that it is indeed running on an SGX enabled machine. During the mbedTLS handshake the client  will first  verify server's certificates (available at Examples/aks-attestation/certs/), after that the server will receive the client's X.509 certificate. This X.509 certificate has SGX quote embedded into it. This SGX quote is verified by the server using DCAP quote verification libraries. Thanks.

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: 9 of 13 files reviewed, 19 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Examples/aks-attestation/README.md, line 32 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskGHW80BCI3NlmipDO:-MtH5ol-8klsghgevyYx:b-wm5xw1) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L32)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

@boryspoplawski Here, the client wants to prove to the server that it is indeed running on an SGX enabled machine. During the mbedTLS handshake the client  will first  verify server's certificates (available at Examples/aks-attestation/certs/), after that the server will receive the client's X.509 certificate. This X.509 certificate has SGX quote embedded into it. This SGX quote is verified by the server using DCAP quote verification libraries. Thanks.
</blockquote></details>

We can modify this example to have server run in a separate cluster itself (to imply that both are in separate machines/infra), making SGX attestation relevant. Is this OK?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: 9 of 13 files reviewed, 19 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Examples/aks-attestation/README.md, line 32 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskGHW80BCI3NlmipDO:-MtKs1rnEqz18Z5g1_Sb:b4bd7dv) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L32)):*
> We can modify this example to have server run in a separate cluster 

It doesn't change anything. Either you trust the cloud provider and do not need SGX, or do not and then you need to verify the process from a trusted machine. A cluster cannot be trusted and untrusted both at the same time, it wouldn't make sense.

aneessahib (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: 9 of 13 files reviewed, 19 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Examples/aks-attestation/README.md, line 32 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskGHW80BCI3NlmipDO:-MtM0RLH8R3Lq2MuOx9S:b1sbpva) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L32)):*
<details><summary><i>Previously, boryspoplawski (Borys Popławski) wrote…</i></summary><blockquote>

> We can modify this example to have server run in a separate cluster 

It doesn't change anything. Either you trust the cloud provider and do not need SGX, or do not and then you need to verify the process from a trusted machine. A cluster cannot be trusted and untrusted both at the same time, it wouldn't make sense.
</blockquote></details>

What we are saying is that the verifier can run in a separate cluster hosted in one of our local systems - hence making it a trusted cluster.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: 9 of 13 files reviewed, 19 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Examples/aks-attestation/README.md, line 32 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskGHW80BCI3NlmipDO:-MtMIqJqCo1n4DJXParp:b2fh9df) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L32)):*

What we are saying is that the verifier can run in a separate cluster hosted in one of our local systems - hence making it a trusted cluster. 
</blockquote></details>

All right, then that verifier doesn't have to run in SGX - there is no need for that since the cluster is trusted.

mkow (changes requested):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @boryspoplawski, @dimakuv, @mkow, and @veenasai2)

___
*[Examples/aks-attestation/aks-secret-prov-client-deployment.yaml, line 22 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msjs88J1ntC6eCymPUZ:-MtbrHxSAZpzcxeska4A:bvistgg) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-client-deployment.yaml#L22)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

this domain name is assigned to the container where ra-tls-secret-prov server is running. We declare the domain name in aks-secret-prov-server-deployment.yaml file as  ``` service.beta.kubernetes.io/azure-dns-label-name: ra-tls-server-aks-dns``` 
</blockquote></details>

But here you hardcoded some domain, which I guess you own? We don't know where the server will be running for the user who tries this example.
___
*[Examples/aks-attestation/aks-secret-prov-server.dockerfile, line 16 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-Msk8quS3H9N3WXBNklL:-Mtbrc938j7lyZB5BnB7:b-s9owfw) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/aks-secret-prov-server.dockerfile#L16)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

In the past we noticed some regression with latest az-dcap-client versions. That's why we kept the one that is working for us. Thanks.
</blockquote></details>

Could you check if that's still the case? We really shouldn't hardcode some old versions of software (new ones may include security fixes).
___
*[Examples/aks-attestation/base-image-generation-script.sh, line 6 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskCoRODvDlv5A_8PW9:-Mtbs27k49lmoLwqWmVO:bdequd7) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/base-image-generation-script.sh#L6)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

First step is to create ra-tls-secret-prov libraries which are later used for creating  base client and server images. Second step is where we graminize client image as part of gsc. Hence, being of different version will not impact each other. Thanks.
</blockquote></details>

@dimakuv: Is this ok? I don't think we'll keep ra-tls compatible between versions? (i.e. that you can use some old client + new server)
Why can't we take the libraries from the image which has Gramine built inside? AFAIR there was a step where GSC built Gramine in a separate image and only afterwards added the user workload, creating a separate image.
___
*[Examples/aks-attestation/gramine_build.sh, line 2 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskA-f9AM_PNkmi0Omc:-MtbsnzkAm59P0RNG4Ex:b9ww2bt) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/gramine_build.sh#L2)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

done, thanks.
</blockquote></details>

But now it points to the part of the docs _after_ discussing the kernel driver and its installation, but in the sentence above you mention the driver (the context suggests that the link will explain in more detail the driver requirements).
___
*[Examples/aks-attestation/README.md, line 28 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskB1rf7MbgnQSJOoxY:-MtbuEFyAwRL_MccyBuJ:beic16w) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L28)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

This domain name is user configured domain name for his deployment. Depending upon the Azure subscription this domain could be anything (e.g. eastus, westus) , that's why we kept it open for the user to fill. User just need to provid the AKS-DNS-NAME in server deployment file and the complete dns name can be fetched from the azure portal post deployment.
</blockquote></details>

Yup, but that's not what I'm asking about, the question is whether someone else can get `AKS-DNS-NAME.SOMETHING_DIFFERENT_HERE.cloudapp.azure.com`.
___
*[Examples/aks-attestation/README.md, line 32 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskGHW80BCI3NlmipDO:-Mtbv2DHEA7FoU9xnY2e:b-lb042y) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L32)):*
<details><summary><i>Previously, boryspoplawski (Borys Popławski) wrote…</i></summary><blockquote>

All right, then that verifier doesn't have to run in SGX - there is no need for that since the cluster is trusted.
</blockquote></details>

+1 to what Borys says, what's the point of running the server under Gramine if we don't even attest its enclave?
___
*[Examples/aks-attestation/README.md, line 54 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskHH020iGfz0-BZXUH:-MtbvF_H90pVJKLYbW0D:bw9da0f) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L54)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

This is not a code snippet. This is just a reference deployment file that the user can customise.
</blockquote></details>

But why is this a list of points, with just one point? It reads quite weirdly for me.
___
*[Examples/aks-attestation/README.md, line 124 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskJ8uSDgvRc2-X490d:-MtbvZcQ9Ria8YAAhkPI:bs33fjj) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L124)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

Step 1: Server (not necessarily running on SGX enabled machine): Waiting for connections. It has a domain name:port as  ra-tls-server-aks-dns.eastus.cloudapp.azure.com:4433

Step 2: Client (Running on SGX enabled machine): Sends its SGX quote to the above server.

After verifying the quote, the server will provision a dummy secret to the client. 

 
</blockquote></details>

Could you make it more clear in the readme that this is the model presented here? I think it's the opposite of the standard one (where the enclaves are the servers providing some services).
Or maybe just reverse it and use the more common model?
___
*[Examples/aks-attestation/certs/README, line 8 at r1](https://reviewable.io/reviews/gramineproject/gsc/38#-MtbwEvYCch35JGbHC9Z:-MtbwEvYCch35JGbHC9_:bvdzmmy) ([raw file](https://github.com/gramineproject/gsc/blob/c3162d513e45381a47567ef96481d301fe6ec26d/Examples/aks-attestation/certs/README#L8)):*
> ```
>   The "Common Name" field is set to `ra-tls-server-aks-dns.eastus.cloudapp.azure.com`.
> ```

I don't understand this. How can this work for anyone except you? This is a domain which you (as I assume) have ownership of. How can someone following this tutorial host anything there?
___
*[Examples/aks-attestation/certs/README, line 8 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskKIss9XeIMNZEJxDz:-Mtbw-1k1IWTt8DaNpMt:by6hcc1) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/certs/README#L8)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

This is just a sample certificate to get the demo working. We are explicitely calling out the fields the user would have to modify before generating their actual production certificates. 
</blockquote></details>

Could we instead generate the certificate during build? This way we won't risk someone deploying this dummy certificate anywhere.
___
*[Examples/aks-attestation/certs/README, line 9 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskKXL7DjaYGjhEVoTU:-Mtbw_vO3PpXNC-iUBwZ:b-fj6sf5) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/certs/README#L9)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

Actually here, I am following the same certficates directory format that we have for https://github.com/gramineproject/gramine/tree/master/CI-Examples/ra-tls-secret-prov/certs . Your point is valid. Here, we are providing the key just to reduce an extra step to the sample working. I can call out seperately in README that the key should not be used in production. 
</blockquote></details>

Why not generate the key during build?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @dimakuv, @mkow, and @veenasai2)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/38#-MtcB2ZM0MCzdYAO0ubl:-MtcEq3a2_jk6wypRAcN:b-nseaqa) (no related file):*
>  see that Michal and Borys would prefer a "turn-key" example that would collect all the relevant information (domain names, etc.) and auto-generate the required files (private keys, certificates).

I've not reviewed the whole PR, just some parts - I've not seen the part with domain names and certificates, so I cannot comment on that (including specifying what I "prefer").

---

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @dimakuv, @mkow, and @veenasai2)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/38#-MtcB2ZM0MCzdYAO0ubl:-MtcNIBa7c5RU_huej1Y:bd6lkjb) (no related file):*
> From the discussions, I see that Michal and Borys would prefer a "turn-key" example that would collect all the relevant information (domain names, etc.) and auto-generate the required files (private keys, certificates).

That would be better, but even without this, the problem is that there are some magic hardcoded DNS names which has to be replaced by the user without being marked as such. Let's at least put something like "YOUR_CLUSTER_NAME" there.

> The approach of this PR is more of a tutorial: look how we did it, and adjust our scripts to your parameters (like domain name) and your needs (running the Secret Provisioning server on your local laptop instead of the AKS cluster).

This point from my side is about suggesting a deployment which makes no sense. I'm fine with a simplistic example, but it should emulate a real case. Here we're e.g. putting both sides into enclaves, but attest only one. Put yourself into shoes of someone who's just learning everything here (that's why they read this example) and they see this. I'd be totally confused.

> One solution to this I can think of: what about moving this AKS tutorial to https://github.com/gramineproject/contrib?

For contrib I think the current state would be ok, but definitely not for the official examples.

---
___
*[Examples/aks-attestation/base-image-generation-script.sh, line 6 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskCoRODvDlv5A_8PW9:-MtcOXhc5is0B6Ozeefu:b89iztm) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/base-image-generation-script.sh#L6)):*

> Why can't we take the libraries from the image which has Gramine built inside? AFAIR there was a step where GSC built Gramine in a separate image and only afterwards added the user workload, creating a separate image.

Hm, yes, but we don't have a way to create a DCAP-enabled Gramine image via `gsc build` currently: https://github.com/gramineproject/gsc/blob/master/templates/Dockerfile.ubuntu.compile.template#L45-L54

IIRC, this PR was started long before we finalized the `meson setup -Ddcap=enabled` compilation option. If I would start from scratch, I would first create a PR to the GSC core that adds a new GSC build option like `gsc build --enable-dcap`. This would change the GSC build process of Gramine to enable all the DCAP stuff, including RA-TLS and Secret Provisioning libraries. Then I believe we would not need these bash scripts in this PR.

> I don't think we'll keep ra-tls compatible between versions? (i.e. that you can use some old client + new server)

Ideally, we should (and we currently do). These are shared libraries -- they are supposed to have backwards-compatible APIs.
</blockquote></details>

Hmm, then I don't like that we're working on a tutorial which does things in an already-deprecated way ;)
How much work would be to add `--enable-dcap` to GSC? Seems pretty straightforward and probably also needed by other usecases.
___
*[Examples/aks-attestation/README.md, line 124 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskJ8uSDgvRc2-X490d:-MtcOsZj48JrhaPFiG0S:b-6lsxl4) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L124)):*

> Or maybe just reverse it and use the more common model?

No, this would be super-confusing. Actually, I tried to look at Kubernetes (secret provisioning) and MS Azure (KeyVault) documents but didn't find any good easy-to-follow terminology.

> Could you make it more clear in the readme that this is the model presented here? I think it's the opposite of the standard one (where the enclaves are the servers providing some services).

We have a Secret Provisioning scenario here: a centralized service (Secret Provisioning Server) runs forever and waits for remote systems to connect to it (clients of Secret Provisioning). This service has the secrets loaded into it (passwords, certificates, etc.). Upon client requests, this service validates the client attestation evidence (the SGX quote) and if validation succeeds, this service releases the secret to the client. The clients in this scenario are SGX enclaves.

So the simplest thing to change in this README is to make all mentions of "servers" and "clients" more specific:
- `server` -> `Secret Provisioning service`
- `client` -> `SGX application (client of the Secret Provisioning service)`
</blockquote></details>

But what's the real-world case for such a thing? I'd rather imagine that the server wants to outsource part of the work to the cloud, so it spawns workers in enclaves in the cloud and connect to them. Not that the workers are spawned (by who, actually?) and they connect to the server.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @dimakuv, @mkow, and @veenasai2)

___
*[Examples/aks-attestation/base-image-generation-script.sh, line 6 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskCoRODvDlv5A_8PW9:-MtcRAOs8hkoPIjXjfT_:b-ufxb60) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/base-image-generation-script.sh#L6)):*

Hmm, then I don't like that we're working on a tutorial which does things in an already-deprecated way ;)
How much work would be to add `--enable-dcap` to GSC? Seems pretty straightforward and probably also needed by other usecases.
</blockquote></details>

I agree that adding `--enable-dcap` to GSC is the way to go. We definitely should implement this feature, independent of this PR. And then this PR can build on top of that DCAP-enabling one.
___
*[Examples/aks-attestation/README.md, line 124 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskJ8uSDgvRc2-X490d:-MtcRM_R07rIZ1ju34Hw:b-6we95e) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L124)):*

But what's the real-world case for such a thing? I'd rather imagine that the server wants to outsource part of the work to the cloud, so it spawns workers in enclaves in the cloud and connect to them. Not that the workers are spawned (by who, actually?) and they connect to the server.
</blockquote></details>

Actually, many cloud deployments will be like this. A good example is Marblerun: https://gramine.readthedocs.io/en/latest/attestation.html#edgeless-marblerun. There is one central entity (the Coordinator), which serves as both the enclave spawner and the secret provisioning service. In case of Marblerun, the Coordinator == the server, and Marbles (SGX enclaves) == clients of this server.

This is not the only example. From what I understand, Microsoft wants to do the same with its MAA (Microsoft Azure Attestation) infrastructure.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @dimakuv, @mkow, and @veenasai2)

___
*[Examples/aks-attestation/README.md, line 124 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskJ8uSDgvRc2-X490d:-MtcT5uBA8eWRypqMgWZ:b77xp34) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L124)):*
> which serves as both the enclave spawner and the secret provisioning service. In case of Marblerun, the Coordinator == the server, and Marbles (SGX enclaves) == clients of this server.

But it's actually the Coordinator connecting and sending data to the workers. Maybe it's just that client-server naming doesn't work well here and we should use coordinator-workers instead?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @dimakuv, @mkow, and @veenasai2)

___
*[Examples/aks-attestation/README.md, line 124 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskJ8uSDgvRc2-X490d:-MtclceIBLVSpcNHac9o:bn53ns3) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L124)):*
> But it's actually the Coordinator connecting and sending data to the workers.

Not exactly. In Marblerun (in case of Gramine), the Coordinator generates envvars like `env.MARBLERUN_COORDINATOR_IP_ADDRESS = "1.1.1.1"`, puts them in the manifest files, transfers this binaries+manifest bundle to the other machine and kicks Gramine to start on that other machine. Then Gramine starts the pre-main logic of the SGX enclave that looks at envvars and connects to the Coordinator. So I would still call it a classic server-client architecture, where SGX enclaves are clients of Coordinator.

> Maybe it's just that client-server naming doesn't work well here and we should use coordinator-workers instead?

I would prefer to have a renaming like this then:
- server -> Secret Provisioning service
- client -> SGX application (or enclavized application, or graminized application)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/38)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @dimakuv, @mkow, and @veenasai2)

___
*[Examples/aks-attestation/README.md, line 124 at r2](https://reviewable.io/reviews/gramineproject/gsc/38#-MskJ8uSDgvRc2-X490d:-MtnEh-09uyOYEKkel-K:bacxgen) ([raw file](https://github.com/gramineproject/gsc/blob/48100050781f5adc73fefd6a2bae1925d2f39521/Examples/aks-attestation/README.md#L124)):*
> I would prefer to have a renaming like this then: [...]

sgtm

Issue #30, opened 2021-11-09
Support for non-root user with GSC docker container


sivashax:
Support for non-root user with GSC docker build is needed. Currently support is available only for ROOT user.

## Steps to reproduce
1. copy the attached docker file into <Gramine Directory>/gsc/test folder.
2. Below command to build the docker.
	a. cd <Gramine Directory>/gsc
	b. docker build --tag ubuntu18.04-bash-nonroot --file test/ubuntu18.04-bash_nonroot.dockerfile .
	c. ./gsc build --insecure-args ubuntu18.04-bash-nonroot test/ubuntu18.04-bash.manifest
	d. ./gsc sign-image ubuntu18.04-bash-nonroot enclave-key.pem

## Expected results
Build is successful.

## Actual results
Below is the error seen with GSC build.
```
E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
E: Unable to lock directory /var/lib/apt/lists/
Failed to build unsigned graminized docker image `gsc-ubuntu18.04-bash-nonroot-unsigned`.
```

sivashax:
[ubuntu18.04-bash-nonroot.txt](https://github.com/gramineproject/gsc/files/7504810/ubuntu18.04-bash-nonroot.txt)
Attached docker file

aneessahib:
assign to me

PR #60, opened 2022-05-25
Allow normal user to run GSC images

veenasai2:
Signed-off-by: “Veena <veena.saini@intel.com>

## Description of the changes 

Currently, gsc does not support building app images that has a non-root user.

The gsc build step fails at line https://github.com/gramineproject/gsc/blob/master/templates/Dockerfile.common.build.template#L11  , with the error as shown below

_E: List directory /var/lib/apt/lists/partial is missing. - Acquire (2: No such file or directory)_

This error occurs because distro-specific packages cannot be installed as non-root user.

If one bypass this error by putting "USER root" at https://github.com/gramineproject/gsc/blob/master/templates/Dockerfile.common.build.template#L10 , then the unsigned image and the final signed gsc image will have user as "root".

This PR allows to create gsc images for app images with non-root user. Also, the original "user" of the app image will be restored back in the final signed gsc image.

## How to test this PR? 
Step 1: `docker pull <non-root-user-app-image>`   //example of non-root user image [bitnami pytorch](https://github.com/bitnami/bitnami-docker-pytorch/blob/master/1/debian-10/Dockerfile#L28)
Step 2: `docker inspect non-root-user-app-image`  // here you can check the original user
Step 3: perform gsc build and gsc sign on the _non-root-user-app-image_
Step 4: `docker inspect gsc-non-root-user-app-image`  // Verify again, the image will have the original user
Step 5: `docker run --device=/dev/sgx_enclave -it gsc-non-root-user-app-image`  //success

Step 6: gsc build and sign should pass for root user images as well.

**Note:** If the app image has an empty user field like user : "" , then the final gsc image will be created with root user.

### DIFF
--- a/finalize_manifest.py
+++ b/finalize_manifest.py
@@ -107,7 +107,7 @@ def main(args=None):
     env = jinja2.Environment(loader=jinja2.FileSystemLoader('/'))
     env.globals.update({'library_paths': generate_library_paths(), 'env_path': os.getenv('PATH')})
 
-    manifest = '/entrypoint.manifest'
+    manifest = '/gramine/app_files/entrypoint.manifest'
     rendered_manifest = env.get_template(manifest).render()
     rendered_manifest_dict = toml.loads(rendered_manifest)
     already_added_files = extract_files_from_user_manifest(rendered_manifest_dict)
--- a/gsc.py
+++ b/gsc.py
@@ -124,6 +124,12 @@ def extract_build_args(args):
                 sys.exit(1)
     return buildargs_dict
 
+def extract_user_from_image_config(config):
+    user = config['User']
+    if not user:
+        user = "root"
+    return user
+
 def merge_two_dicts(dict1, dict2, path=[]):
     for key in dict2:
         if key in dict1:
@@ -167,6 +173,8 @@ def gsc_build(args):
     env.globals.update(yaml.safe_load(args.config_file))
     env.globals.update(vars(args))
     env.globals.update({'app_image': original_image_name})
+    original_image_user = extract_user_from_image_config(original_image.attrs['Config'])
+    env.globals.update({'app_user': original_image_user})
     extract_binary_cmd_from_image_config(original_image.attrs['Config'], env)
     extract_working_dir_from_image_config(original_image.attrs['Config'], env)
 
--- a/templates/Dockerfile.common.build.template
+++ b/templates/Dockerfile.common.build.template
@@ -8,6 +8,11 @@ FROM gsc-{{Gramine.Image}} AS gramine
 # Combine Gramine image with the original app image
 FROM {{app_image}}
 
+# App image "user" is stored, and execution will continue as root to install packages.
+# In the signed graminized image, the "user" will be restored.
+ENV app_user={{app_user}}
+USER root
+
 # Install distro-specific packages to run Gramine (e.g., python3, protobuf, toml, etc.)
 {% block install %}{% endblock %}
 
@@ -18,10 +23,19 @@ COPY --from=gramine /gramine/ /gramine/
 COPY --from=gramine /gramine/meson_build_output /gramine/meson_build_output
 {% endif %}
 
+# Create a directory that will store apploader.sh and entrypoint files.
+RUN mkdir -p /gramine/app_files
+
+# Make the app image user owner of /gramine/app_files directory
+RUN chown $app_user /gramine/app_files/
+
+# Grant the owner of the directory (i.e., the app image user here) all the permissions
+RUN chmod u+rwx /gramine/app_files/
+
 # Copy helper scripts and Gramine manifest
 COPY *.py /
-COPY apploader.sh /
-COPY entrypoint.manifest /
+COPY apploader.sh /gramine/app_files/
+COPY entrypoint.manifest /gramine/app_files/
 
 # Generate trusted arguments if required
 {% if not insecure_args %}
@@ -37,10 +51,10 @@ RUN cd / \
 ENV PATH="/gramine/meson_build_output/bin:$PATH"
 
 # Mark apploader.sh executable, finalize manifest, and remove intermediate scripts
-RUN chmod u+x /apploader.sh \
+RUN chmod u+x /gramine/app_files/apploader.sh \
     && /usr/bin/python3 -B /finalize_manifest.py \
     && rm -f /finalize_manifest.py
 
 # Define default command
-ENTRYPOINT ["/bin/bash", "/apploader.sh"]
+ENTRYPOINT ["/bin/bash", "/gramine/app_files/apploader.sh"]
 CMD [{% if insecure_args %} "{{'", "'.join(cmd)}}" {% endif %}]
--- a/templates/Dockerfile.common.sign.template
+++ b/templates/Dockerfile.common.sign.template
@@ -1,15 +1,18 @@
 # Sign image in a separate stage to ensure that signing key is never part of the final image
 FROM {{image}} as unsigned_image
 
-COPY gsc-signer-key.pem /gsc-signer-key.pem
+COPY gsc-signer-key.pem /gramine/app_files/gsc-signer-key.pem
 
 RUN {% block path %}{% endblock %} gramine-sgx-sign \
-      --key /gsc-signer-key.pem \
-      --manifest /entrypoint.manifest \
-      --output /entrypoint.manifest.sgx
+      --key /gramine/app_files/gsc-signer-key.pem \
+      --manifest /gramine/app_files/entrypoint.manifest \
+      --output /gramine/app_files/entrypoint.manifest.sgx
 
 # This trick removes all temporary files from the previous commands (including gsc-signer-key.pem)
 FROM {{image}}
 
-COPY --from=unsigned_image /*.sig /
-COPY --from=unsigned_image /*.sgx /
+COPY --from=unsigned_image /gramine/app_files/*.sig /gramine/app_files/
+COPY --from=unsigned_image /gramine/app_files/*.sgx /gramine/app_files/
+
+# Exit root user and restore back the app image user
+USER $app_user
--- a/templates/apploader.common.template
+++ b/templates/apploader.common.template
@@ -8,8 +8,8 @@ set -ex
 # Default to Linux-SGX if no PAL was specified
 if [ -z "$GSC_PAL" ] || [ "$GSC_PAL" == "Linux-SGX" ]
 then
-    gramine-sgx-get-token --sig /entrypoint.sig --output /entrypoint.token
-    gramine-sgx /entrypoint {% if insecure_args %}{{binary_arguments}} "${@}"{% endif %}
+    gramine-sgx-get-token --sig /gramine/app_files/entrypoint.sig --output /gramine/app_files/entrypoint.token
+    gramine-sgx /gramine/app_files/entrypoint {% if insecure_args %}{{binary_arguments}} "${@}"{% endif %}
 else
-    gramine-direct /entrypoint {{binary_arguments}} "${@}"
+    gramine-direct /gramine/app_files/entrypoint {{binary_arguments}} "${@}"
 fi
--- a/test/generic.manifest
+++ b/test/generic.manifest
@@ -5,5 +5,5 @@ sgx.enclave_size = "4G"
 sgx.thread_num = 8
 
 sgx.trusted_files = [
-  "file:/entrypoint.manifest",  # unused entry, only to test merging of manifests
+  "file:/gramine/app_files/entrypoint.manifest",  # unused entry, only to test merging of manifests
 ]


aneessahib:
fixes issue #60

dimakuv (changes requested):
@aneessahib Did you mean #30? Looks like a typo.

> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60#-:-N3KCQVfEAGqz8vYq_8x:b-8jgw9e)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @veenasai2)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/60#-N3KGUs8A7npihoCXhqB):*
Rephrase: `Support Docker images that have a non-root user`

We can fix this during final rebase.
___
*[`gsc.py` line 177 at r1](https://reviewable.io/reviews/gramineproject/gsc/60#-N3KFYHx529bWCshGEXD:-N3KFYHx529bWCshGEXE:b-ptdc0t) ([raw file](https://github.com/gramineproject/gsc/blob/8fe1b9e328780e96e2a20ed13d305dd686a26b07/gsc.py#L177)):*
> ```Python
>     env.globals.update({'app_image': original_image_name})
>     original_image_user = extract_user_from_image_config(original_image.attrs['Config'])
>     env.globals.update({'app_user': original_image_user})
> ```

Could you please change the new function to be similar as already-existing functions like `extract_binary_cmd_from_image_config()`?

In particular, pass the `env` variable into the new function and perform `env.globals.update()` in the function itself.
___
*[`templates/Dockerfile.common.build.template` line 14 at r1](https://reviewable.io/reviews/gramineproject/gsc/60#-N3KFmizD3VgX5J5uAhk:-N3KFmizD3VgX5J5uAhl:b-4xgr1z) ([raw file](https://github.com/gramineproject/gsc/blob/8fe1b9e328780e96e2a20ed13d305dd686a26b07/templates/Dockerfile.common.build.template#L14)):*
> ```
> # In the signed graminized image, the "user" will be restored.
> ENV app_user={{app_user}}
> USER root
> ```

Why so complicated? Why can't you temporarily switch to the root user (like you do now) and then -- immediately after `block install` -- switch back to normal user?

I mean, I propose this:
```
# temporarily switch to the root user to install packages
USER root

# Install distro-specific packages to run Gramine (e.g., python3, protobuf, toml, etc.)
{% block install %}{% endblock %}

# switch back to original app-image user
USER {{app_user}}
```

See also https://stackoverflow.com/a/63643361/6325371

In this case, you won't need any other modifications. In particular, you won't need `/gramine/app_files` changes.

Or did I miss something?

aneessahib (commented):
That's right, sorry.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60#-:-N3KICWy-_JvdBrDC28v:blg9tnd)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @veenasai2)

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: 5 of 6 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/60#-N3KGUs8A7npihoCXhqB):*

Rephrase: `Support Docker images that have a non-root user`

We can fix this during final rebase.
</blockquote></details>

Sure, thanks.
___
*[`gsc.py` line 177 at r1](https://reviewable.io/reviews/gramineproject/gsc/60#-N3KFYHx529bWCshGEXD:-N3Nl1Ja1eK9hBpet4m1:beaonw6) ([raw file](https://github.com/gramineproject/gsc/blob/8fe1b9e328780e96e2a20ed13d305dd686a26b07/gsc.py#L177)):*

Could you please change the new function to be similar as already-existing functions like `extract_binary_cmd_from_image_config()`?

In particular, pass the `env` variable into the new function and perform `env.globals.update()` in the function itself.
</blockquote></details>

done, thanks. 
___
*[`templates/Dockerfile.common.build.template` line 14 at r1](https://reviewable.io/reviews/gramineproject/gsc/60#-N3KFmizD3VgX5J5uAhk:-N3Nl3dk4xdkHDmaet6b:b-vxxgi2) ([raw file](https://github.com/gramineproject/gsc/blob/8fe1b9e328780e96e2a20ed13d305dd686a26b07/templates/Dockerfile.common.build.template#L14)):*

Why so complicated? Why can't you temporarily switch to the root user (like you do now) and then -- immediately after `block install` -- switch back to normal user?

I mean, I propose this:
```
# temporarily switch to the root user to install packages
USER root

# Install distro-specific packages to run Gramine (e.g., python3, protobuf, toml, etc.)
{% block install %}{% endblock %}

# switch back to original app-image user
USER {{app_user}}
```

See also https://stackoverflow.com/a/63643361/6325371

In this case, you won't need any other modifications. In particular, you won't need `/gramine/app_files` changes.

Or did I miss something?
</blockquote></details>

We can't switch back to normal user immediately after block install. As, we will need root permissions further down in the code for something like https://github.com/gramineproject/gsc/blob/master/templates/Dockerfile.common.build.template#L40 

Thanks.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @veenasai2)

___
*[`templates/Dockerfile.common.build.template` line 14 at r1](https://reviewable.io/reviews/gramineproject/gsc/60#-N3KFmizD3VgX5J5uAhk:-N3NsVt60G69tLB8BoqL:b-yjhxxh) ([raw file](https://github.com/gramineproject/gsc/blob/8fe1b9e328780e96e2a20ed13d305dd686a26b07/templates/Dockerfile.common.build.template#L14)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

We can't switch back to normal user immediately after block install. As, we will need root permissions further down in the code for something like https://github.com/gramineproject/gsc/blob/master/templates/Dockerfile.common.build.template#L40 

Thanks.
</blockquote></details>

Hm, why do we need root permissions at that point? What is the error message?

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`templates/Dockerfile.common.build.template` line 14 at r1](https://reviewable.io/reviews/gramineproject/gsc/60#-N3KFmizD3VgX5J5uAhk:-N3OOeMgCloBm8gDWvC5:b-ileub5) ([raw file](https://github.com/gramineproject/gsc/blob/8fe1b9e328780e96e2a20ed13d305dd686a26b07/templates/Dockerfile.common.build.template#L14)):*

Ah, so this is because we're putting all files in the root directory and doing everything in this root directory. Which is not accessible by a non-root user. Makes sense.

So the way you circumvent this is to create a `/gramine/app_files/` directory and re-assign it to the non-root user.

But still, why can't you do these operations (that require the root user) in one code block and then switch back to normal user? something like this:
```
# temporarily switch to the root user to install packages
USER root

# Install distro-specific packages to run Gramine (e.g., python3, protobuf, toml, etc.)
{% block install %}{% endblock %}

# Create a directory that will store apploader.sh and entrypoint files.
RUN mkdir -p /gramine/app_files

# Make the app image user owner of /gramine/app_files directory
RUN chown {{app_user}} /gramine/app_files/

# switch back to original app-image user
USER {{app_user}}
```

This should be enough, no?

My main point with these proposed changes is that I don't like dragging the root user until the very end. I think we should switch to the normal user at the first possible moment.
</blockquote></details>

Yeah, your approach is better. In that case we need to make some additional changes as well , like 

```
COPY --chown={{app_user}} *.py /
COPY --chown={{app_user}} apploader.sh /gramine/app_files/
COPY --chown={{app_user}} entrypoint.manifest /gramine/app_files/
```

otherwise again we will see

`chmod: changing permissions of '/gramine/app_files/apploader.sh': Operation not permitted` .

For now, I am not pushing the suggested changes as we have one more hidden dependency in finalize_manifest.py  (for non-root user) at https://github.com/gramineproject/gsc/blob/master/finalize_manifest.py#L107 and https://github.com/gramineproject/gsc/blob/master/finalize_manifest.py#L85 that translates to an error

` subprocess.CalledProcessError: Command 'ldconfig -v' returned non-zero exit status 1. ` 

I  will resolve this error and push the changes. 

Thanks.

aneessahib (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: 3 of 7 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @veenasai2)

___
*[`templates/entrypoint.common.manifest.template` line 40 at r3](https://reviewable.io/reviews/gramineproject/gsc/60#-N3XyF5OCQSBHJfBHQqD:-N3XyF5OCQSBHJfBHQqE:bhnmz6b) ([raw file](https://github.com/gramineproject/gsc/blob/ba1507a20cb660de86d1e6cdbe36e998ef5f759c/templates/entrypoint.common.manifest.template#L40)):*
> ```
>   "file:/etc/shadow-",
> ]
> ```

So you run into issues during the signing step, since the non root user does not have access to these files. By adding them `allowed_files` in manifest would result in skipping being added as trusted, resulting in a smooth signing stage, but I'm not sure if this is the right usage of `allowed_files`. May be other potential solutions
1) If user does not have access while signing, then ignore just go to the next file. This change would have to go into the sign tool, but this could be considered hacky.
2) Add something like `sgx.skip_file` , that GSC would ignore during the trusted file generation. This could be later used to optimize the size of large GSC images too. 

@dimakuv  what do you think

dimakuv (changes requested):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @dimakuv, and @veenasai2)

___
*[`finalize_manifest.py` line 38 at r3](https://reviewable.io/reviews/gramineproject/gsc/60#-N3YDG9mBnPTn2BGymM6:-N3YDG9mBnPTn2BGymM7:b-tbcjdm) ([raw file](https://github.com/gramineproject/gsc/blob/ba1507a20cb660de86d1e6cdbe36e998ef5f759c/finalize_manifest.py#L38)):*
> ```Python
>                                 r'|\.dockerenv'
>                                 r'|\.dockerinit'
>                                 r'|etc/mtab'
> ```

The new files to ignore should go into this list, not into `sgx.allowed_files` in `template/entrypoint.common.manifest.template`
___
*[`finalize_manifest.py` line 85 at r3](https://reviewable.io/reviews/gramineproject/gsc/60#-N3YE_AT5CHsdrfvKi6E:-N3YE_AT5CHsdrfvKi6F:b-8hybc3) ([raw file](https://github.com/gramineproject/gsc/blob/ba1507a20cb660de86d1e6cdbe36e998ef5f759c/finalize_manifest.py#L85)):*
> ```Python
> def generate_library_paths():
>     encoding = sys.stdout.encoding if sys.stdout.encoding is not None else 'UTF-8'
>     ld_paths = subprocess.check_output('ldconfig -v -N -X', stderr=subprocess.PIPE, shell=True)
> ```

FYI: Explaining this small change for future readers.

The previous version `ldconfig -v` had a side-effect of updating the caches and symlinks of the most recently used shared libraries. But we simply want to read the list of system paths where the libraries can be found; we don't want any side-effects.

So we had this bug previously, and never noticed it, because we only allowed `root` users -- and they have all privileges. So `ldconfig -v` silently updated the system-wide information.

Now that we allow non-root users, we hit this problem because `ldconfig -v` requires root privileges to update the caches. So we added `-N -X` to prevent the side-effects: `-N` forbids to rebuild the cache, `-X` forbids to update links.
___
*[`templates/Dockerfile.common.build.template` line 14 at r1](https://reviewable.io/reviews/gramineproject/gsc/60#-N3KFmizD3VgX5J5uAhk:-N3YEK3S8rNCv9bmPHQL:b20eqiq) ([raw file](https://github.com/gramineproject/gsc/blob/8fe1b9e328780e96e2a20ed13d305dd686a26b07/templates/Dockerfile.common.build.template#L14)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

Yeah, your approach is better. In that case we need to make some additional changes as well , like 

```
COPY --chown={{app_user}} *.py /
COPY --chown={{app_user}} apploader.sh /gramine/app_files/
COPY --chown={{app_user}} entrypoint.manifest /gramine/app_files/
```

otherwise again we will see

`chmod: changing permissions of '/gramine/app_files/apploader.sh': Operation not permitted` .

For now, I am not pushing the suggested changes as we have one more hidden dependency in finalize_manifest.py  (for non-root user) at https://github.com/gramineproject/gsc/blob/master/finalize_manifest.py#L107 and https://github.com/gramineproject/gsc/blob/master/finalize_manifest.py#L85 that translates to an error

` subprocess.CalledProcessError: Command 'ldconfig -v' returned non-zero exit status 1. ` 

I  will resolve this error and push the changes. 

Thanks.


</blockquote></details>

This is resolved, right? Please mark it as "Done".
___
*[`templates/entrypoint.common.manifest.template` line 40 at r3](https://reviewable.io/reviews/gramineproject/gsc/60#-N3XyF5OCQSBHJfBHQqD:-N3YDSQO5kIdyv9HWrgS:b2pa9v) ([raw file](https://github.com/gramineproject/gsc/blob/ba1507a20cb660de86d1e6cdbe36e998ef5f759c/templates/entrypoint.common.manifest.template#L40)):*

So you run into issues during the signing step, since the non root user does not have access to these files. By adding them `allowed_files` in manifest would result in skipping being added as trusted, resulting in a smooth signing stage, but I'm not sure if this is the right usage of `allowed_files`. May be other potential solutions
1) If user does not have access while signing, then ignore just go to the next file. This change would have to go into the sign tool, but this could be considered hacky.
2) Add something like `sgx.skip_file` , that GSC would ignore during the trusted file generation. This could be later used to optimize the size of large GSC images too. 

@dimakuv  what do you think
</blockquote></details>

Yes, we already have the function in our trusted file generation script (`finalize_manifest.py`), so we should use option 2.

Please delete the files from here and add them to the list pointed to by my other comment.

Btw, since that list uses regex, you can replace these:
```
  "file:/etc/gshadow",
  "file:/etc/gshadow-",
```

with a single line like this:
```
r'|etc/gshadow.*'
```

(It means "all files under etc/ that start with gshadow prefix")

aneessahib (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @veenasai2)

___
*[`templates/entrypoint.common.manifest.template` line 40 at r3](https://reviewable.io/reviews/gramineproject/gsc/60#-N3XyF5OCQSBHJfBHQqD:-N3YGxSR3uiMcZK9-1s1:bkr27dy) ([raw file](https://github.com/gramineproject/gsc/blob/ba1507a20cb660de86d1e6cdbe36e998ef5f759c/templates/entrypoint.common.manifest.template#L40)):*

Yes, we already have the function in our trusted file generation script (`finalize_manifest.py`), so we should use option 2.

Please delete the files from here and add them to the list pointed to by my other comment.

Btw, since that list uses regex, you can replace these:
```
  "file:/etc/gshadow",
  "file:/etc/gshadow-",
```

with a single line like this:
```
r'|etc/gshadow.*'
```

(It means "all files under etc/ that start with gshadow prefix")
</blockquote></details>

Yes this had crossed our minds too, but this would remove access for root users as well wouldnt it? Do we want that?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @dimakuv, and @veenasai2)

___
*[`templates/entrypoint.common.manifest.template` line 40 at r3](https://reviewable.io/reviews/gramineproject/gsc/60#-N3XyF5OCQSBHJfBHQqD:-N3YN3r1Eu_L4r_NTKwD:bwyreu7) ([raw file](https://github.com/gramineproject/gsc/blob/ba1507a20cb660de86d1e6cdbe36e998ef5f759c/templates/entrypoint.common.manifest.template#L40)):*

Yes this had crossed our minds too, but this would remove access for root users as well wouldnt it? Do we want that?
</blockquote></details>

Looking at the list of files, they all pertain to passwords/user names. No real-world application that users would want to run in GSC/Gramine needs these files. Moreover, Gramine simply doesn't have any functionality to work with these files.

So yes, we can definitely remove access to these files, even for root users. I see no problems with it.

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: 5 of 7 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib and @dimakuv)

___
*[`finalize_manifest.py` line 38 at r3](https://reviewable.io/reviews/gramineproject/gsc/60#-N3YDG9mBnPTn2BGymM6:-N3YUBc76tgTxn4BXr8a:b-lhs69o) ([raw file](https://github.com/gramineproject/gsc/blob/ba1507a20cb660de86d1e6cdbe36e998ef5f759c/finalize_manifest.py#L38)):*

The new files to ignore should go into this list, not into `sgx.allowed_files` in `template/entrypoint.common.manifest.template`
</blockquote></details>

done, thanks
___
*[`templates/entrypoint.common.manifest.template` line 40 at r3](https://reviewable.io/reviews/gramineproject/gsc/60#-N3XyF5OCQSBHJfBHQqD:-N3YU1LB4RqKsYTcfHsr:b29xy6r) ([raw file](https://github.com/gramineproject/gsc/blob/ba1507a20cb660de86d1e6cdbe36e998ef5f759c/templates/entrypoint.common.manifest.template#L40)):*

Looking at the list of files, they all pertain to passwords/user names. No real-world application that users would want to run in GSC/Gramine needs these files. Moreover, Gramine simply doesn't have any functionality to work with these files.

So yes, we can definitely remove access to these files, even for root users. I see no problems with it.
</blockquote></details>

The suggested changes are done. Thanks.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @dimakuv, and @veenasai2)

___
*[`finalize_manifest.py` line 41 at r4](https://reviewable.io/reviews/gramineproject/gsc/60#-N3YVcGM7BPUYDGCKOdy:-N3YVcGM7BPUYDGCKOdz:b-dnie5l) ([raw file](https://github.com/gramineproject/gsc/blob/11b9537d30fa1b4b0c18959500c127428d225fba/finalize_manifest.py#L41)):*
> ```Python
>                                 r'|etc/gshadow.*'
>                                 r'|etc/mtab'
>                                 r'|etc/.pwd.lock'
> ```

You need to escape dots, so it should look like this:
```
r'|etc/\.pwd\.lock'
```
___
*[`finalize_manifest.py` line 43 at r4](https://reviewable.io/reviews/gramineproject/gsc/60#-N3YVqDgEjmU2tnV7-zt:-N3YVqDh0NhwoX2nEhmA:b5aj5ut) ([raw file](https://github.com/gramineproject/gsc/blob/11b9537d30fa1b4b0c18959500c127428d225fba/finalize_manifest.py#L43)):*
> ```Python
>                                 r'|etc/.pwd.lock'
>                                 r'|etc/rc(\d|.)\.d/.*'
>                                 r'|etc/security/opasswd'
> ```

Actually, I think you should just skip the whole directory:
```
r'|etc/security/.*'
```

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @dimakuv, and @veenasai2)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/60#-N3YbFVFCDjtl6-RvJ2h:-N3YbFVI2os-r80Ie8VK:b7gtnyb) (no related file):*
The PR looks good to me. Could you verify on 2-3 examples (one with non-root user, the others classic with root user) this final version?

___

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @aneessahib and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/60#-N3YbFVFCDjtl6-RvJ2h:-N3YcEBF2IfAnLKEdRQV:bhtism5) (no related file):*

The PR looks good to me. Could you verify on 2-3 examples (one with non-root user, the others classic with root user) this final version?
</blockquote></details>

@dimakuv, I have verified this final version on (1) python image from Docker hub (root user), (2) bash image (gsc/test) (root user),  (3) hello-world image (gsc/test) with explicitly putting non-root user, (4) bitnami-pytorch image from docker hub (non-root user).  Thanks.

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @aneessahib and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/60#-N3YbFVFCDjtl6-RvJ2h:-N3Z-Kb7-pZIznAHEna9:bend12m) (no related file):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

@dimakuv, I have verified this final version on (1) python image from Docker hub (root user), (2) bash image (gsc/test) (root user),  (3) hello-world image (gsc/test) with explicitly putting non-root user, (4) bitnami-pytorch image from docker hub (non-root user).  Thanks.
</blockquote></details>

Nice, thanks for verification!

___

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: 5 of 7 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @dimakuv, and @mkow)

___
*[`gsc.py` line 130 at r5](https://reviewable.io/reviews/gramineproject/gsc/60#-N4TCxTv9RZ_Nfeaf4qV:-N4VYiMe9521hXknNYMH:b-r18wie) ([raw file](https://github.com/gramineproject/gsc/blob/c22f7c18215c3d0d4dfa75364d08acece4469cab/gsc.py#L130)):*
> ```Python
>     user = config['User']
>     if not user:
>         user = "root"
> ```

done, thanks.
___
*[`templates/Dockerfile.common.build.template` line 11 at r5](https://reviewable.io/reviews/gramineproject/gsc/60#-N4U2ue10rb4uU3bTJkm:-N4VYkPT9XHXbNVZChvy:b-uy3xr9) ([raw file](https://github.com/gramineproject/gsc/blob/c22f7c18215c3d0d4dfa75364d08acece4469cab/templates/Dockerfile.common.build.template#L11)):*

Why is this in doublequotes?
</blockquote></details>

I wanted to emphasize the user here. But anyways, I removed the entire comment as the env variable is no longer used. Thanks.
___
*[`templates/Dockerfile.common.build.template` line 13 at r5](https://reviewable.io/reviews/gramineproject/gsc/60#-N4U4wQM3iV9iMdiyQ01:-N4VZ6286yPGtDBtIDxm:b-feytoi) ([raw file](https://github.com/gramineproject/gsc/blob/c22f7c18215c3d0d4dfa75364d08acece4469cab/templates/Dockerfile.common.build.template#L13)):*

Are you even using this env variable anywhere?
</blockquote></details>

No , we are not using this anymore. In the initial proposal I was using it.  Thanks.
___
*[`templates/Dockerfile.common.build.template` line 13 at r5](https://reviewable.io/reviews/gramineproject/gsc/60#-N4U4Zo922n57ZIvez2G:-N4VZFeI4WTUp4PPcB1o:b-bq11mn) ([raw file](https://github.com/gramineproject/gsc/blob/c22f7c18215c3d0d4dfa75364d08acece4469cab/templates/Dockerfile.common.build.template#L13)):*
> ```
> # App image "user" is stored, and execution will continue as root to install packages.
> # In the signed graminized image, the "user" will be restored.
> ENV app_user={{app_user}}
> ```

Removed the app_user env variable. Thanks.

dimakuv (commented):
> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @dimakuv, and @mkow)

mkow (approved):
> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @aneessahib and @dimakuv)

___
*[`templates/Dockerfile.common.build.template` line 11 at r5](https://reviewable.io/reviews/gramineproject/gsc/60#-N4U2ue10rb4uU3bTJkm:-N4Wx8qj6Pj3kiPPnvrG:b7mx4rm) ([raw file](https://github.com/gramineproject/gsc/blob/c22f7c18215c3d0d4dfa75364d08acece4469cab/templates/Dockerfile.common.build.template#L11)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

I wanted to emphasize the user here. But anyways, I removed the entire comment as the env variable is no longer used. Thanks.
</blockquote></details>

Doublequotes are used to cite something, for emphasis you should use `*` or `_`.

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @aneessahib and @dimakuv)

___
*[`templates/Dockerfile.common.build.template` line 11 at r5](https://reviewable.io/reviews/gramineproject/gsc/60#-N4U2ue10rb4uU3bTJkm:-N4X0XoF3_fTZ4yXEGMH:b-jcyzg0) ([raw file](https://github.com/gramineproject/gsc/blob/c22f7c18215c3d0d4dfa75364d08acece4469cab/templates/Dockerfile.common.build.template#L11)):*

Doublequotes are used to cite something, for emphasis you should use `*` or `_`.
</blockquote></details>

ok, thanks.

dimakuv (approved):
> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @aneessahib)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/60#-N3KGUs8A7npihoCXhqB):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

Sure, thanks.
</blockquote></details>

Done
___
*[`gsc.py` line 395 at r7](https://reviewable.io/reviews/gramineproject/gsc/60#-N4b_dmk6A8EC_-1xsL-:-N4b_dmk6A8EC_-1xsL0:b-sav8cs) ([raw file](https://github.com/gramineproject/gsc/blob/62b2d9b8d4f9978970a0f37245c35e564980e66d/gsc.py#L395)):*
> ```Python
>         # Copy sigstruct file from Docker container into temporary directory on the host
>         docker_socket.containers.run(args.image,
>             '\'cp /gramine/app_files/entrypoint.sig /tmp/host/ 2>/dev/null || :\'',
> ```

FYI: Noticed this small bug when manually testing (with `./gsc info-image gsc-python`).

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @aneessahib and @veenasai2)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/60#-N4b_pJ_6zJIqX1J442i:-N4b_pJ_6zJIqX1J442j:b-cubmnc) (no related file):*
@veenasai2 This PR is ready for merge. I tested it on a couple classic images (Python, Bash). Could you also test it on your Bitnami Pytorch image, to be absolutely sure? If your test is successful, then we merge.

___

mkow (dismissed):
> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @aneessahib and @veenasai2)

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @aneessahib and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/60#-N4b_pJ_6zJIqX1J442i:-N4byiDPCQTmg9Swbsvc:b-j1aa8j) (no related file):*

@veenasai2 This PR is ready for merge. I tested it on a couple classic images (Python, Bash). Could you also test it on your Bitnami Pytorch image, to be absolutely sure? If your test is successful, then we merge.
</blockquote></details>

@dimakuv , Just now I tested the PR on Bitnami Pytorch image that has a non-root user. The PR is working fine, except the entrypoint.sig part. 

___
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/60#-N3KGUs8A7npihoCXhqB):*

Done
</blockquote></details>

Thanks
___
*[`gsc.py` line 395 at r7](https://reviewable.io/reviews/gramineproject/gsc/60#-N4b_dmk6A8EC_-1xsL-:-N4bzaoJ79kx0PGOFkNO:b-8x5b98) ([raw file](https://github.com/gramineproject/gsc/blob/62b2d9b8d4f9978970a0f37245c35e564980e66d/gsc.py#L395)):*

FYI: Noticed this small bug when manually testing (with `./gsc info-image gsc-python`).
</blockquote></details>

The path for the entrypoint.sig needs to be updated. I will upload the changes soon. Thanks.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @aneessahib and @dimakuv)

___
*[`gsc.py` line 395 at r7](https://reviewable.io/reviews/gramineproject/gsc/60#-N4b_dmk6A8EC_-1xsL-:-N4c6vt784LQ-f3I0uTl:b-64r5nj) ([raw file](https://github.com/gramineproject/gsc/blob/62b2d9b8d4f9978970a0f37245c35e564980e66d/gsc.py#L395)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

The path for the entrypoint.sig needs to be updated. I will upload the changes soon. Thanks.
</blockquote></details>

@veenasai2 Wait, but I fixed it. What do you mean? Are you sure you used my latest version? I updated it just today. so that it works.

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @aneessahib and @dimakuv)

___
*[`gsc.py` line 395 at r7](https://reviewable.io/reviews/gramineproject/gsc/60#-N4b_dmk6A8EC_-1xsL-:-N4cCEas3n_ZnVDiawRB:bdecrcb) ([raw file](https://github.com/gramineproject/gsc/blob/62b2d9b8d4f9978970a0f37245c35e564980e66d/gsc.py#L395)):*

@veenasai2 Wait, but I fixed it. What do you mean? Are you sure you used my latest version? I updated it just today. so that it works.
</blockquote></details>

@dimakuv I tried with your commit now, getting the below error for the bitnami pytorch image.

 with open(os.path.join(tmpdirname, "entrypoint.sig"), 'rb') as sig:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpvwz8smtw/entrypoint.sig' 

However,  with `./gsc info-image gsc-python` , no errors. Thanks

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @aneessahib and @dimakuv)

___
*[`gsc.py` line 395 at r7](https://reviewable.io/reviews/gramineproject/gsc/60#-N4b_dmk6A8EC_-1xsL-:-N4cKawh-MuzAtIqR2NH:b-fgy8s4) ([raw file](https://github.com/gramineproject/gsc/blob/62b2d9b8d4f9978970a0f37245c35e564980e66d/gsc.py#L395)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

@dimakuv I tried with your commit now, getting the below error for the bitnami pytorch image.

 with open(os.path.join(tmpdirname, "entrypoint.sig"), 'rb') as sig:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpvwz8smtw/entrypoint.sig' 

However,  with `./gsc info-image gsc-python` , no errors. Thanks

</blockquote></details>

What is producing the error? Which command? Can you copy the whole output? I don't get it, you first say that there is an error, but then you say that there is no error. So which one is it?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @aneessahib, @dimakuv, and @veenasai2)

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @aneessahib and @dimakuv)

___
*[`gsc.py` line 395 at r7](https://reviewable.io/reviews/gramineproject/gsc/60#-N4b_dmk6A8EC_-1xsL-:-N4cMBLnEsAonODIYp13:bbfcjia) ([raw file](https://github.com/gramineproject/gsc/blob/62b2d9b8d4f9978970a0f37245c35e564980e66d/gsc.py#L395)):*

What is producing the error? Which command? Can you copy the whole output? I don't get it, you first say that there is an error, but then you say that there is no error. So which one is it?
</blockquote></details>

``` 
./gsc info-image gsc-base-bitnami-image
Traceback (most recent call last):
  File "./gsc", line 12, in <module>
    sys.exit(main(sys.argv))
  File "./gsc.py", line 481, in main
    return args.command(args)
  File "./gsc.py", line 399, in gsc_info_image
    with open(os.path.join(tmpdirname, "entrypoint.sig"), 'rb') as sig:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp90t3j4q1/entrypoint.sig'

```

```
 ./gsc info-image gsc-python
mr_enclave = "XXXXXXXXXXXXXXXXXX"
mr_signer = "XXXXXXXXXXXXXXXXXXX"
isv_prod_id = X
isv_svn = X
date = "2022-06-15"
flags = "XXXXXXXXXXXXX"
xfrms = "XXXXXXXXXXXXXX"
misc_select = "XXXXXXXXXXXXX"
debug = false

```

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @aneessahib and @dimakuv)

___
*[`gsc.py` line 395 at r7](https://reviewable.io/reviews/gramineproject/gsc/60#-N4b_dmk6A8EC_-1xsL-:-N4cPMsv9ZdaY4R1UrFv:b-z5ffc1) ([raw file](https://github.com/gramineproject/gsc/blob/62b2d9b8d4f9978970a0f37245c35e564980e66d/gsc.py#L395)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

``` 
./gsc info-image gsc-base-bitnami-image
Traceback (most recent call last):
  File "./gsc", line 12, in <module>
    sys.exit(main(sys.argv))
  File "./gsc.py", line 481, in main
    return args.command(args)
  File "./gsc.py", line 399, in gsc_info_image
    with open(os.path.join(tmpdirname, "entrypoint.sig"), 'rb') as sig:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp90t3j4q1/entrypoint.sig'

```

```
 ./gsc info-image gsc-python
mr_enclave = "XXXXXXXXXXXXXXXXXX"
mr_signer = "XXXXXXXXXXXXXXXXXXX"
isv_prod_id = X
isv_svn = X
date = "2022-06-15"
flags = "XXXXXXXXXXXXX"
xfrms = "XXXXXXXXXXXXXX"
misc_select = "XXXXXXXXXXXXX"
debug = false

```
</blockquote></details>

```
From ubuntu:18.04

RUN apt-get update

CMD ["bash"]

USER 1001

```

if we convert the bash image ( which is in gsc/test folder) to non-root user image (let say  bash-non-root-image )
, we can see this issue happening.

```
./gsc info-image gsc-bash-non-root-image
Traceback (most recent call last):
  File "./gsc", line 12, in <module>
    sys.exit(main(sys.argv))
  File "./gsc.py", line 481, in main
    return args.command(args)
  File "./gsc.py", line 399, in gsc_info_image
    with open(os.path.join(tmpdirname, "entrypoint.sig"), 'rb') as sig:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp3mm3aocl/entrypoint.sig'

```

Basically, with non-root user image, I am seeing this issue.

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: 6 of 7 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib, @dimakuv, and @mkow)

___
*[`gsc.py` line 395 at r7](https://reviewable.io/reviews/gramineproject/gsc/60#-N4b_dmk6A8EC_-1xsL-:-N4k5esq6VR1gtXCJ14-:betoau6) ([raw file](https://github.com/gramineproject/gsc/blob/62b2d9b8d4f9978970a0f37245c35e564980e66d/gsc.py#L395)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

```
From ubuntu:18.04

RUN apt-get update

CMD ["bash"]

USER 1001

```

if we convert the bash image ( which is in gsc/test folder) to non-root user image (let say  bash-non-root-image )
, we can see this issue happening.

```
./gsc info-image gsc-bash-non-root-image
Traceback (most recent call last):
  File "./gsc", line 12, in <module>
    sys.exit(main(sys.argv))
  File "./gsc.py", line 481, in main
    return args.command(args)
  File "./gsc.py", line 399, in gsc_info_image
    with open(os.path.join(tmpdirname, "entrypoint.sig"), 'rb') as sig:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp3mm3aocl/entrypoint.sig'

```

Basically, with non-root user image, I am seeing this issue. 


</blockquote></details>

@dimakuv This issue is fixed now. I tested the new version with python image (root user) and bitnami image (non-root user). Thanks.

dimakuv (changes requested):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @aneessahib and @veenasai2)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/60#-N4b_pJ_6zJIqX1J442i:-N4kNw7OELbLzO-NhUN0:b7n21ya) (no related file):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

@dimakuv , Just now I tested the PR on Bitnami Pytorch image that has a non-root user. The PR is working fine, except the entrypoint.sig part. 
</blockquote></details>

OK, `./gsc info-image` on a non-root-user Docker image works now.

___
___
*[`gsc.py` line 395 at r7](https://reviewable.io/reviews/gramineproject/gsc/60#-N4b_dmk6A8EC_-1xsL-:-N4kMsoU9I9JrDBjKHBX:b-2t65in) ([raw file](https://github.com/gramineproject/gsc/blob/62b2d9b8d4f9978970a0f37245c35e564980e66d/gsc.py#L395)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

@dimakuv This issue is fixed now. I tested the new version with python image (root user) and bitnami image (non-root user). Thanks.
</blockquote></details>

For the context: the issue stems from three facts:
- Python's `TemporaryDirectory()` creates a temporary directory with permissions 644 (owner can read and write, everyone else can only read).
- The root-user Docker image operates under root and can write to any directory anyway.
- The non-root-user Docker image operates under a UID that can be different from the UID of the temporary directory (in this case, Docker user is `1001` and host user is `1000`). So the Docker user **cannot** write to the directory owned by the host user. Error happens.

The fix proposed by @veenasai2 is to allow *everyone* to write into the directory. Since this is a temporary directory created solely for this purpose, and is deleted afterwards, I see no problems with this approach.
___
*[`gsc.py` line 394 at r8](https://reviewable.io/reviews/gramineproject/gsc/60#-N4kLy1T4IGhRZtMnvCA:-N4kLy1T4IGhRZtMnvCB:b-c9qafq) ([raw file](https://github.com/gramineproject/gsc/blob/bd9d524f3be03e2d78c88da813df937fa893c1db/gsc.py#L394)):*
> ```Python
>     with tempfile.TemporaryDirectory() as tmpdirname:
>         # Grant temporary directory all the permissions for non-root user image access
>         os.chmod(tmpdirname,0o777)
> ```

This should be `666` (we don't allow executing files). I will fix it during the final rebase. And we'll need to test it then again.

dimakuv (approved):
> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @aneessahib)

___
*[`gsc.py` line 394 at r8](https://reviewable.io/reviews/gramineproject/gsc/60#-N4kLy1T4IGhRZtMnvCA:-N4kPlCw8ksXq2QT0wSs:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/bd9d524f3be03e2d78c88da813df937fa893c1db/gsc.py#L394)):*

This should be `666` (we don't allow executing files). I will fix it during the final rebase. And we'll need to test it then again.
</blockquote></details>

Done.

veenasai2 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @aneessahib)

___
*[`gsc.py` line 394 at r8](https://reviewable.io/reviews/gramineproject/gsc/60#-N4kLy1T4IGhRZtMnvCA:-N4ldL9p6ebp_CA-nMIy:b-4pdhj2) ([raw file](https://github.com/gramineproject/gsc/blob/bd9d524f3be03e2d78c88da813df937fa893c1db/gsc.py#L394)):*

Done.
</blockquote></details>

@dimakuv, thanks for pushing these changes.  I tested the latest version, here are the results:

   `./gsc info-image gsc-python` gives following error with 666 (though works fine 777): 

```
  File "./gsc.py", line 404, in gsc_info_image
    with open(os.path.join(tmpdirname, "entrypoint.sig"), 'rb') as sig:
PermissionError: [Errno 13] Permission denied: '/tmp/tmpy1tza3yr/entrypoint.sig'

```

same error , I got for bitnami image too. 

I will do more analysis on why this is happening and get back soon. Thanks

veenasai2:
> > _[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)_ status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @aneessahib)
> 
> _[`gsc.py` line 394 at r8](https://reviewable.io/reviews/gramineproject/gsc/60#-N4kLy1T4IGhRZtMnvCA:-N4ldL9p6ebp_CA-nMIy:b-4pdhj2) ([raw file](https://github.com/gramineproject/gsc/blob/bd9d524f3be03e2d78c88da813df937fa893c1db/gsc.py#L394)):_
> 
> Previously, dimakuv (Dmitrii Kuvaiskii) wrote…
> @dimakuv, thanks for pushing these changes. I tested the latest version, here are the results:
> 
> `./gsc info-image gsc-python` gives following error with 666 (though works fine 777):
> 
> ```
>   File "./gsc.py", line 404, in gsc_info_image
>     with open(os.path.join(tmpdirname, "entrypoint.sig"), 'rb') as sig:
> PermissionError: [Errno 13] Permission denied: '/tmp/tmpy1tza3yr/entrypoint.sig'
> ```
> 
> same error , I got for bitnami image too.
> 
> I will do more analysis on why this is happening and get back soon. Thanks

Based on [this answer](https://unix.stackexchange.com/questions/21251/execute-vs-read-bit-how-do-directory-permissions-in-linux-work) (3rd point) , looks like we need execute bit too set in order to enter into a directory and read a file.

dimakuv (approved):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @aneessahib)

___
*[`gsc.py` line 394 at r8](https://reviewable.io/reviews/gramineproject/gsc/60#-N4kLy1T4IGhRZtMnvCA:-N4zrZwq7wv_-PYWaZsh:bdln9ey) ([raw file](https://github.com/gramineproject/gsc/blob/bd9d524f3be03e2d78c88da813df937fa893c1db/gsc.py#L394)):*
<details><summary><i>Previously, veenasai2 wrote…</i></summary><blockquote>

Based on [this answer](https://unix.stackexchange.com/questions/21251/execute-vs-read-bit-how-do-directory-permissions-in-linux-work) (3rd point) , looks like we need execute bit too set in order to enter into a directory and read a file.
</blockquote></details>

Oops, you are right. Done, reverted back to your `777`.

I tested on the classic Python image, it works fine. @veenasai2 @please test on your non-root-user image, and we should finally merge this PR.

mkow (approved):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: all files reviewed, 1 unresolved discussion (waiting on @aneessahib)

dimakuv (approved):
> Dismissed @aneessahib from a discussion.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/60)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #62, opened 2022-06-01
Allow to skip SGX token generation

dimakuv:
## Description of the changes 

SGX Launch Token (aka EINITTOKEN) is generated right-before Gramine launch via the `gramine-sgx-get-token` tool. For DCAP platforms, the SGX token is unused, so there is no need to generate it. This commit introduces a GSC switch `--skip-token-generation` for this.

This is useful for DCAP platforms (one less file in the Docker container), and enables such scenarios as `docker run --read-only`.
Also, skipping running `gramine-sgx-get-token` paves a path to removing Python as a run-time dependency for GSC-built Docker images.

Related PR in core Gramine: https://github.com/gramineproject/gramine/pull/363

## How to test this PR? 

A Python example from docs with `./gsc build --skip-token-generation --insecure-args python test/generic.manifest`. You'll see that `gramine-sgx-get-token` tool is not called during the last step (`docker run`).

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gsc/62)

### DIFF
--- a/Documentation/index.rst
+++ b/Documentation/index.rst
@@ -101,6 +101,11 @@ Gramine image.
    Allow untrusted arguments to be specified at :command:`docker run`. Otherwise
    any arguments specified during :command:`docker run` are ignored.
 
+.. option:: --skip-token-generation
+
+   Allows to skip SGX token generation (via :command:`gramine-sgx-get-token`).
+   Useful for DCAP platforms and read-only filesystems.
+
 .. option:: --no-cache
 
    Disable Docker's caches during :command:`gsc build`. This builds the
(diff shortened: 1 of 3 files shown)


mythi:
I tested this fix and it works well!

FWIW, the resulting `apploader.sh` looks like this but that's probably fine :-)

```
then

    gramine-sgx /entrypoint  "${@}"
else
```

dimakuv:
> FWIW, the resulting `apploader.sh` looks like this but that's probably fine :-)

I think we can live with it :) Btw, we'll need to remove `set -x` from the apploader script; it's annoying to see all those logs "hey user, look what I'm executing now".

dimakuv:
Closing this PR as it is considered as a stop-gap solution, and also it is not high priority. We better fix it properly.

See https://github.com/gramineproject/gramine/pull/363 for details.

PR #109, opened 2022-11-02
Drop support for out-of-tree DCAP SGX drivers

dimakuv:
## Description of the changes 

The out-of-tree DCAP drivers have the same userspace interface as the upstream driver, so we can re-use the flow for upstream one.

This is a counterpart PR to these ones for core Gramine:
- https://github.com/gramineproject/gramine/pull/997
- https://github.com/gramineproject/gramine/pull/1027

Note that GSC will still be able to run with older Gramine (e.g. v1.3.1). This is because GSC now uses `-Dsgx_driver=upstream`, so older Gramine will recognize this.

**Note that the reverse is false**: one cannot use old GSC (before this PR) with the new Gramine, if the system has an oot DCAP driver -- in this case GSC will set `Dsgx_driver=dcap1.10` which is not known to new Gramine.

Finally, note that if GSC is used with an old `config.yaml` file with something like:
```
SGXDriver:
    Repository: "https://github.com/intel/SGXDataCenterAttestationPrimitives.git"
    Branch:     "DCAP_1.10.3 && cp -r driver/linux/* ."
```
this will still work, because GSC will simply ignore this (oot DCAP driver) repo and default to `upstream`.

## How to test this PR? 

Manually on different machines with different `config.yaml` configs: different `SGXDriver` and different `Gramine.Branch`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gsc/109)

### DIFF
--- a/config.yaml.template
+++ b/config.yaml.template
@@ -21,17 +21,17 @@ Gramine:
 # Specify the Intel SGX driver installed on your machine (more specifically, on the machine where
 # the graminized Docker container will run); there are several variants of the SGX driver:
 #
+#   - upstream (in-kernel) driver: use empty values like below
+#         Repository: ""
+#         Branch:     ""
+#
+#   - DCAP out-of-tree driver: same as above, use empty values
+#         Repository: ""
+#         Branch:     ""
+#
 #   - legacy out-of-tree driver: use something like the below values, but adjust the branch name
 #         Repository: "https://github.com/01org/linux-sgx-driver.git"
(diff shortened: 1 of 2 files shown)


mkow (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/109)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #112, opened 2022-12-09
Enable `--template` option to use with sign-image (for use with HSMs,for example)

svenkata9:
Signed-off-by: Sankaranarayanan Venkatasubramanian <sankaranarayanan.venkatasubramanian@intel.com>

## Description
The `gsc` tool cannot do production signing on the gramized docker images today, and this PR enables that. This PR enables passing a 'self-contained' Dockerfile as an argument to `gsc sign-image` with the option `--template` (newly introduced in this PR).  For example, using a template like [this ](https://github.com/svenkata9/contrib/blob/svenkat9/akv_sign/Integrations/azure/akv-sign/Dockerfile.sign.akv.debian.template), we can do production signing with Azure Key Vault. Once this PR is merged, `contrib` repo will be updated with templates for Azure Key Vault and OpenSSL engine.

## How to test this PR? 
For testing this PR, you will need to use the Dockerfile template mentioned above for use with AKV, or create one based on this for using local signing. and then you will be able to sign the unsigned gsc image using `./gsc sign-image --template Dockerfile.sign.akv.debian.template`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ __pycache__
 *.manifest.sgx
 *.sig
 *.token
+/templates/Dockerfile.sign.user.template
--- a/gsc.py
+++ b/gsc.py
@@ -350,26 +350,36 @@ def gsc_sign_image(args):
 
     distro, _ = distro.split(':')
     env.loader = jinja2.FileSystemLoader('templates/')
-    sign_template = env.get_template(f'{distro}/Dockerfile.sign.template')
+    sign_template = []
+    build_args = []
 
     os.makedirs(tmp_build_path, exist_ok=True)
+
+    # Use default steps if user has not provided a Dockerfile/template for signing
+    if args.template is None:
+        # copy user-provided signing key and signing Bash script to our tmp build dir (to copy them
+        # later inside Docker image)
+        tmp_build_key_path = tmp_build_path / 'gsc-signer-key.pem'
+        tmp_build_sign_path = tmp_build_path / 'sign.sh'
+        shutil.copyfile(os.path.abspath(args.key), tmp_build_key_path)
+        shutil.copy(os.path.abspath('sign.sh'), tmp_build_sign_path)
+        sign_template = env.get_template(f'{distro}/Dockerfile.sign.template')
+        build_args = {"passphrase": args.passphrase}
+    else:
+        shutil.copy(args.template, 'templates/Dockerfile.sign.user.template')
+        sign_template = env.get_template(f'{distro}/Dockerfile.sign.user.template')
+
     with open(tmp_build_path / 'Dockerfile.sign', 'w') as dockerfile:
         dockerfile.write(sign_template.render(image=unsigned_image_name))
-
-    # copy user-provided signing key and signing Bash script to our tmp build dir (to copy them
-    # later inside Docker image)
-    tmp_build_key_path = tmp_build_path / 'gsc-signer-key.pem'
-    tmp_build_sign_path = tmp_build_path / 'sign.sh'
-    shutil.copyfile(os.path.abspath(args.key), tmp_build_key_path)
-    shutil.copy(os.path.abspath('sign.sh'), tmp_build_sign_path)
 
     try:
         # `forcerm` parameter forces removal of intermediate Docker images even after unsuccessful
         # builds, to not leave the signing key lingering in any Docker containers
         build_docker_image(docker_socket.api, tmp_build_path, signed_image_name, 'Dockerfile.sign',
-                           forcerm=True, buildargs={"passphrase": args.passphrase})
+                           forcerm=True, buildargs=build_args)
     finally:
-        os.remove(tmp_build_key_path)
+        if args.template is None:
+            os.remove(tmp_build_key_path)
 
     if get_docker_image(docker_socket, signed_image_name) is None:
         print(f'Failed to build a signed graminized Docker image `{signed_image_name}`.')
@@ -501,7 +511,9 @@ sub_sign.set_defaults(command=gsc_sign_image)
 sub_sign.add_argument('-c', '--config_file', type=argparse.FileType('r', encoding='UTF-8'),
     default='config.yaml', help='Specify configuration file.')
 sub_sign.add_argument('image', help='Name of the application (base) Docker image.')
-sub_sign.add_argument('key', help='Key to sign the Intel SGX enclaves inside the Docker image.')
+sub_sign.add_argument('-k', '--key', help='Key to sign the Intel SGX enclaves inside the Docker image.')
+sub_sign.add_argument('-t', '--template',
+                      help='Custom Dockerfile/template to use for signing, say, with a HSM.')
 sub_sign.add_argument('-p', '--passphrase', help='Passphrase for the signing key.')
 
 sub_info = subcommands.add_parser('info-image', help='Retrieve information about a graminized '
--- a/templates/centos/Dockerfile.sign.user.template
+++ b/templates/centos/Dockerfile.sign.user.template
@@ -1,0 +1,3 @@
+{% extends "Dockerfile.sign.user.template" %}
+
+{% block path %}export PYTHONPATH="${PYTHONPATH}:$(find /gramine/meson_build_output/lib64 -type d -path '*/site-packages')" &&{% endblock %}
--- a/templates/debian/Dockerfile.sign.user.template
+++ b/templates/debian/Dockerfile.sign.user.template
@@ -1,0 +1,3 @@
+{% extends "Dockerfile.sign.user.template" %}
+
+{% block path %}export PYTHONPATH="${PYTHONPATH}:$(find /gramine/meson_build_output/lib -type d -path '*/site-packages')" &&{% endblock %}
--- a/templates/ubuntu/Dockerfile.sign.user.template
+++ b/templates/ubuntu/Dockerfile.sign.user.template
@@ -1,0 +1,2 @@
+{% extends "debian/Dockerfile.sign.user.template" %}
+


dimakuv (changes requested):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/112)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @svenkata9)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrBkWfAy5gY_aVHye7:-NKrBkWfAy5gY_aVHye8:bk32plz) (no related file):*
You need to update the docs:
1. https://gramine.readthedocs.io/projects/gsc/en/latest/#gsc-sign-image-signs-a-graminized-image -- add `--template` option
2. https://gramine.readthedocs.io/projects/gsc/en/latest/#stages-of-building-graminized-sgx-docker-images -- in stage 3, mention that users can also supply their own Dockerfile templates, to e.g. sign with HSM flows.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrEnlT6yTbP7oefeI0:-NKrEnlT6yTbP7oefeI1:b1cyxtv) (no related file):*
I'm not sure if `--template` is the best name for this option. Maybe `--custom-dockerfile`? I don't know, I can't come up with a better name. So not blocking.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrF2TjCgikd4k5R9C-:-NKrF2TjCgikd4k5R9C0:b-2ycg6k) (no related file):*
The user needs to manually modify the templates, e.g. specify the AKV URL: https://github.com/svenkata9/contrib/blob/54ee9fc3aa1ba367bf203f769c51b555a8310576/Integrations/azure/akv-sign/Dockerfile.sign.akv.debian.template#L23

It would be better to provide such names (e.g. AKV URL) in the `gsc sign-image` command itself. But on the other hand, I understand that this won't be extensible and only confusing... 

___
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrC53c258_Z2GqUJ4O):*
This commit title is too long. Let's split into two: title and message body:
```
Add `--template` option to use with `sign-image` command

Currently `gsc sign-image` command only provides one way to
sign images -- using Gramine's default `gramine-sgx-sign` tool;
this logic is encapsulated in `Dockerfile.common.sign.template`.

This commit adds the ability to sign images using user-specified
tools, e.g. for signing with hardware HSMs provided by Azure KeyVault
and OpenSSL engine.
```
___
*[`.gitignore` line 24 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrD1Uj3pc-rAsAF45c:-NKrD1Uj3pc-rAsAF45d:bxqaj5r) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/.gitignore#L24)):*
> ```ignore
> *.sig
> *.token
> /templates/Dockerfile.sign.user.template
> ```

This doesn't look right. I suggest:
1. Add the removal of this file in `gsc.py` (in the `finally` clause)
2. Remove this line from here.
___
*[`gsc.py` line 353 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrDD4C-Gjw74GKp58N:-NKrDD4C-Gjw74GKp58O:bdxk6cu) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/gsc.py#L353)):*
> ```Python
>     distro, _ = distro.split(':')
>     env.loader = jinja2.FileSystemLoader('templates/')
>     sign_template = []
> ```

This should be `sign_template = None`
___
*[`gsc.py` line 354 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrDFeJ88kkOu_BCrH4:-NKrDFeJ88kkOu_BCrH5:b-zfhwvx) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/gsc.py#L354)):*
> ```Python
>     env.loader = jinja2.FileSystemLoader('templates/')
>     sign_template = []
>     build_args = []
> ```

This should be a dict, not a list: `build_args = {}`
___
*[`gsc.py` line 370 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrDx5T6FxBhOITpvyx:-NKrDx5T6FxBhOITpvyy:b-fdogev) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/gsc.py#L370)):*
> ```Python
>     else:
>         shutil.copy(args.template, 'templates/Dockerfile.sign.user.template')
>         sign_template = env.get_template(f'{distro}/Dockerfile.sign.user.template')
> ```

Actually, don't you want to propagate the `key` into the user template:
```
        build_args = {"key": args.key}
```

this way users won't need to manually modify this line in the template: https://github.com/svenkata9/contrib/blob/54ee9fc3aa1ba367bf203f769c51b555a8310576/Integrations/azure/akv-sign/Dockerfile.sign.akv.debian.template#L24

See the example with `passphrase` usage here: https://github.com/gramineproject/gsc/blob/b7f0576219321c1b7d94b29661274419b009fc49/templates/Dockerfile.common.sign.template#L5
___
*[`gsc.py` line 382 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrEP320h-a3m7XrjtF:-NKrEP320h-a3m7XrjtG:bt32l3e) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/gsc.py#L382)):*
> ```Python
>     finally:
>         if args.template is None:
>             os.remove(tmp_build_key_path)
> ```

Please add the removal of the copied `templates/Dockerfile.sign.user.template`, smth like:
```
    else:
        os.remove('templates/Dockerfile.sign.user.template')
```
___
*[`gsc.py` line 516 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrEdJHA-Y0KCQ8Qivf:-NKrEdJHA-Y0KCQ8Qivg:bvgfeh5) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/gsc.py#L516)):*
> ```Python
> sub_sign.add_argument('-k', '--key', help='Key to sign the Intel SGX enclaves inside the Docker image.')
> sub_sign.add_argument('-t', '--template',
>                       help='Custom Dockerfile/template to use for signing, say, with a HSM.')
> ```

`say, with a HSM` -> `e.g. with an HSM`
___
*[`templates/ubuntu/Dockerfile.sign.user.template` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrElNnD-F9T27ciAl7:-NKrElNnD-F9T27ciAl8:b6sfg8g) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/templates/ubuntu/Dockerfile.sign.user.template#L2)):*
> ```
> {% extends "debian/Dockerfile.sign.user.template" %}
> ```

Please remove this empty line.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/112)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @svenkata9)

___
*[`templates/centos/Dockerfile.sign.user.template` line 1 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrNb5u6GFqK9E8Nmea:-NKrNb5v5uckRtU6OIPt:b5wugj5) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/templates/centos/Dockerfile.sign.user.template#L1)):*
> ```
> {% extends "Dockerfile.sign.user.template" %}
> ```

As I mentioned in https://github.com/gramineproject/contrib/pull/20, I don't think you need these three new template files at all.

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/112)* status: 3 of 6 files reviewed, 12 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrEnlT6yTbP7oefeI0:-NKwAM9bD91HZH3C7u14:b-mpon9q) (no related file):*

I'm not sure if `--template` is the best name for this option. Maybe `--custom-dockerfile`? I don't know, I can't come up with a better name. So not blocking.
</blockquote></details>

I did consider `--dockerfile`, `--custom-script`, `--script` and few others, but since we use all Dockerfiles as templates, I settled on this.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrF2TjCgikd4k5R9C-:-NKwAvyL-9KkD6JO8Gov:b-4amzic) (no related file):*

The user needs to manually modify the templates, e.g. specify the AKV URL: https://github.com/svenkata9/contrib/blob/54ee9fc3aa1ba367bf203f769c51b555a8310576/Integrations/azure/akv-sign/Dockerfile.sign.akv.debian.template#L23

It would be better to provide such names (e.g. AKV URL) in the `gsc sign-image` command itself. But on the other hand, I understand that this won't be extensible and only confusing... 
</blockquote></details>

Yes, the idea is to have the sample templates (in contrib) and have users modify according to their needs by filling in the URL and key name. If we provide options like `--url`, then we will need to add stuff like `--engine` for OpenSSL engine-based signing and we will end up with lot many confusing options. So, to keep it simple, it is `--template <Dockerfile>` that contains all the information.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrBkWfAy5gY_aVHye7:-NKwVPKe1-FUBfnwxFZm:b-896fix) (no related file):*

You need to update the docs:
1. https://gramine.readthedocs.io/projects/gsc/en/latest/#gsc-sign-image-signs-a-graminized-image -- add `--template` option
2. https://gramine.readthedocs.io/projects/gsc/en/latest/#stages-of-building-graminized-sgx-docker-images -- in stage 3, mention that users can also supply their own Dockerfile templates, to e.g. sign with HSM flows.
</blockquote></details>

Done.

___
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrC53c258_Z2GqUJ4O):*

This commit title is too long. Let's split into two: title and message body:
```
Add `--template` option to use with `sign-image` command

Currently `gsc sign-image` command only provides one way to
sign images -- using Gramine's default `gramine-sgx-sign` tool;
this logic is encapsulated in `Dockerfile.common.sign.template`.

This commit adds the ability to sign images using user-specified
tools, e.g. for signing with hardware HSMs provided by Azure KeyVault
and OpenSSL engine.
```
</blockquote></details>

ACK.
___
*[`gsc.py` line 353 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrDD4C-Gjw74GKp58N:-NKwBvmTAdkrUqYMfewi:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/gsc.py#L353)):*

This should be `sign_template = None`
</blockquote></details>

Done.
___
*[`gsc.py` line 354 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrDFeJ88kkOu_BCrH4:-NKwC5R33xM1U1KKdH0z:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/gsc.py#L354)):*

This should be a dict, not a list: `build_args = {}`
</blockquote></details>

Done.
___
*[`gsc.py` line 370 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrDx5T6FxBhOITpvyx:-NKwC7NhCCEVl3etnMQS:bntufyu) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/gsc.py#L370)):*

Actually, don't you want to propagate the `key` into the user template:
```
        build_args = {"key": args.key}
```

this way users won't need to manually modify this line in the template: https://github.com/svenkata9/contrib/blob/54ee9fc3aa1ba367bf203f769c51b555a8310576/Integrations/azure/akv-sign/Dockerfile.sign.akv.debian.template#L24

See the example with `passphrase` usage here: https://github.com/gramineproject/gsc/blob/b7f0576219321c1b7d94b29661274419b009fc49/templates/Dockerfile.common.sign.template#L5
</blockquote></details>

Like I mentioned above, the idea is for the user to provide a Dockerfile as template that contains all the information including key URL and name.
___
*[`gsc.py` line 382 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrEP320h-a3m7XrjtF:-NKwCHx-ErOna0XOXDsb:bxlly3x) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/gsc.py#L382)):*

Please add the removal of the copied `templates/Dockerfile.sign.user.template`, smth like:
```
    else:
        os.remove('templates/Dockerfile.sign.user.template')
```
</blockquote></details>

Not valid comment anymore since we removed the user template concept and consuming the Dockerfile as is.
___
*[`gsc.py` line 516 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrEdJHA-Y0KCQ8Qivf:-NKwCVnu-13Y8Mq1QNBu:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/gsc.py#L516)):*

`say, with a HSM` -> `e.g. with an HSM`
</blockquote></details>

Done.
___
*[`templates/centos/Dockerfile.sign.user.template` line 1 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrNb5u6GFqK9E8Nmea:-NKwCdz47yUDG8IaKEKi:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/templates/centos/Dockerfile.sign.user.template#L1)):*

As I mentioned in https://github.com/gramineproject/contrib/pull/20, I don't think you need these three new template files at all.
</blockquote></details>

Done.
___
*[`templates/ubuntu/Dockerfile.sign.user.template` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrElNnD-F9T27ciAl7:-NKwCgcBEhzxRCYrvDI9:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/templates/ubuntu/Dockerfile.sign.user.template#L2)):*

Please remove this empty line.
</blockquote></details>

Done.
___
*[`.gitignore` line 24 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrD1Uj3pc-rAsAF45c:-NKwBec62Mfgm0tnlhEP:b-rmakoq) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/.gitignore#L24)):*

This doesn't look right. I suggest:
1. Add the removal of this file in `gsc.py` (in the `finally` clause)
2. Remove this line from here.
</blockquote></details>

Eliminated the need for Dockerfile.sign.user.template (and the extensions for OS flavors). So, we can mark this comment as resolved.

dimakuv (changes requested):
> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/112)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @svenkata9)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrEnlT6yTbP7oefeI0:-NL0HuK37z6C89e9LGR3:bx03r28) (no related file):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

I did consider `--dockerfile`, `--custom-script`, `--script` and few others, but since we use all Dockerfiles as templates, I settled on this.
</blockquote></details>

Yes, I can't come up with anything better, so I'm fine.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrF2TjCgikd4k5R9C-:-NL0I0Th1nStrNmFnWmd:bv3fm3y) (no related file):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

Yes, the idea is to have the sample templates (in contrib) and have users modify according to their needs by filling in the URL and key name. If we provide options like `--url`, then we will need to add stuff like `--engine` for OpenSSL engine-based signing and we will end up with lot many confusing options. So, to keep it simple, it is `--template <Dockerfile>` that contains all the information.
</blockquote></details>

Ok, I agree.

___
___
*[`gsc.py` line 370 at r1](https://reviewable.io/reviews/gramineproject/gsc/112#-NKrDx5T6FxBhOITpvyx:-NL0IJGW1mq2v5a7q6KX:bhyg3fz) ([raw file](https://github.com/gramineproject/gsc/blob/e8e5e8de1d38c4944ce3842a49381ebb4c692709/gsc.py#L370)):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

Like I mentioned above, the idea is for the user to provide a Dockerfile as template that contains all the information including key URL and name.
</blockquote></details>

Yes, ok, I agree
___
*[`gsc.py` line 518 at r2](https://reviewable.io/reviews/gramineproject/gsc/112#-NL0Inqa424mDglt8fYk:-NL0Inqa424mDglt8fYl:b-sxb5at) ([raw file](https://github.com/gramineproject/gsc/blob/5c830ffc079d3e0ea3eacb1c652b5b60babc598b/gsc.py#L518)):*
> ```Python
>     help='Key to sign the Intel SGX enclaves inside the Docker image.')
> sub_sign.add_argument('-t', '--template',
>                       help='Custom Dockerfile/template to use for signing, say, with an HSM.')
> ```

Please unify the indentation similar to how it is in other cases (4 spaces from beginning of line)
___
*[`Documentation/index.rst` line 148 at r2](https://reviewable.io/reviews/gramineproject/gsc/112#-NL0JClG4BjaoV-6Zyr6:-NL0JClG4BjaoV-6Zyr7:b-vknetv) ([raw file](https://github.com/gramineproject/gsc/blob/5c830ffc079d3e0ea3eacb1c652b5b60babc598b/Documentation/index.rst#L148)):*
> ```reStructuredText
>    Specify configuration file. Default: :file:`config.yaml`
> 
> .. option:: -k KEY-FILE
> ```

No need to specify this `KEY-FILE` here. Looking at other option descriptions, we don't use such format (`-option name`).
___
*[`Documentation/index.rst` line 158 at r2](https://reviewable.io/reviews/gramineproject/gsc/112#-NL0KHty6b4UpYtGo2Ju:-NL0KHtzC4eZN7lzXW1Z:bbp43pt) ([raw file](https://github.com/gramineproject/gsc/blob/5c830ffc079d3e0ea3eacb1c652b5b60babc598b/Documentation/index.rst#L158)):*
> ```reStructuredText
> .. option:: -t
> 
>    Provide Dockerfile to use for signing the enclave (e.g., with an HSM)
> ```

Please also add a second sentence:
```
If this option is used, then `-k` and `-p` are ignored (that's because
the enclave signing key must be specified in the provided Dockerfile).
```
___
*[`Documentation/index.rst` line 283 at r2](https://reviewable.io/reviews/gramineproject/gsc/112#-NL0JrqP0aMFMLpAlNAv:-NL0JrqP0aMFMLpAlNAw:beixb2l) ([raw file](https://github.com/gramineproject/gsc/blob/5c830ffc079d3e0ea3eacb1c652b5b60babc598b/Documentation/index.rst#L283)):*
> ```reStructuredText
>    provided by the user via the :command:`gsc sign-image` command and copied
>    into this Docker build stage. The users can also supply their own Dockerfile
>    templates using `--template` option of this command to sign (e.g. with HSM).
> ```

Slight rewording:
```
   into this Docker build stage. The users can also supply their own Dockerfile
   templates using `-t`/`--template` command-line option for user-specific
   signining flows (e.g., signing with an HSM).

```

svenkata9:
@woju Please add your comments here and your proposal. I feel that is not solving the problem that we are trying to solve. There are multiple options that I investigated including passing arguments to `gsc` `sign-image`, but that will make it more complicated. Theis `--template` is based on the existing template that we use for local signing and introduces only the HSM specific stuff.

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/112)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @svenkata9)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/112#-NOxuuRf9tE_D_MPmcZ4:-NOxuuRf9tE_D_MPmcZ5:bfbdsi8) (no related file):*
I object to `--template` option as it stands, because:
- GSC integrity should not rely on external template. It takes one clueless developer in the future to remove this line, and GSC would still "work" apart from the small detail that it ships private key used for signing. See comment in gramineproject/contrib#20
- the user shouldn't be required to know GSC implenentation details (i.e. not only this `FROM` device, but also `COPY` paths, `PYTHONPATH` etc.) and in any case should not rely on those details being stable (i.e. we should be able to change anything there without breaking user's overrides.

Argument that "it's us who will do this" and/or "this will live in our repositories" is invalid, because people will take this file as an example, will copy it elsewhere and adapt to their infrastructure.

Secondary consideration is, I don't want to favour any particular HSM or cloud unless they provide resources for CI in the project. Otherwise we can't test this, and by carrying this in our repo, there's an implicit expectation that we don't break it.

So I see two ways forward:
1. we fit azure signing into signing plugins (gramineproject/gramine#1118) (iiuc this means that we add dependencies for those other azure-specific modules and include `subprocess` calls to `az login` and `logout`) and provide `--requirements` option that would install this plugin (from either git+http:// or local repo, as supported by pip), then passthrough the arguments to `gramine-sgx-sign` similarly to `$password` or so
2. if akv signing doesn't fit into signing plugins, then this needs to be rewritten that only parts of the template are changed, but the overall structure is not to be rewritten by the user

One possible way to do the second option would be to have `--template-dir` that would add an override `FileSystemLoader` like this:

```python
loader_gsc_templates = jinja2.FileSystemLoader('templates')
env = jinja2.Environment(loader=jinja2.ChoiceLoader([
    jinja2.PrefixLoader({'': loader_gsc_templates, }, '!'),
    jinja2.FileSystemLoader(user_provided_templates_path),
    loader_gsc_templates,
]))
```

Then you can instrument this signing templates with `{% block %}`s and inherit from the original templates:

```jinja
{% extends '!debian/sign.dockerfile' %}
{% block post-install %}
{{ super() }}
RUN python3 -m pip install whatever
RUN apt-get install also-whatever
{% endblock %}
```

```jinja
{% extends '!rhel/sign.dockerfile' %}
{% block post-install %}
{{ super() }}
RUN python3 -m pip install whatever
RUN yum install also-whatever-but-rpm
{% endblock %}
```

```jinja
{% extends '!sign.dockerfile' %}
{% block pre-sign %}
RUN az login
{% endblock %}

{% block post-sign %}
RUN az logout
{% endblock %}
```

But this would require reorganising the templates to allow not repeating ourselves, rethink names of templates and blocks, document those names, care about backwards compatibility etc.

PS 1b if support for encrypted keys is required, gramineproject/gramine#1197 is an RFC

___

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/112)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/112#-NOxuuRf9tE_D_MPmcZ4:-NOyM6tk3NGEzBXezSG9:ba4weuy) (no related file):*

I object to `--template` option as it stands, because:
- GSC integrity should not rely on external template. It takes one clueless developer in the future to remove this line, and GSC would still "work" apart from the small detail that it ships private key used for signing. See comment in gramineproject/contrib#20
- the user shouldn't be required to know GSC implenentation details (i.e. not only this `FROM` device, but also `COPY` paths, `PYTHONPATH` etc.) and in any case should not rely on those details being stable (i.e. we should be able to change anything there without breaking user's overrides.

Argument that "it's us who will do this" and/or "this will live in our repositories" is invalid, because people will take this file as an example, will copy it elsewhere and adapt to their infrastructure.

Secondary consideration is, I don't want to favour any particular HSM or cloud unless they provide resources for CI in the project. Otherwise we can't test this, and by carrying this in our repo, there's an implicit expectation that we don't break it.

So I see two ways forward:
1. we fit azure signing into signing plugins (gramineproject/gramine#1118) (iiuc this means that we add dependencies for those other azure-specific modules and include `subprocess` calls to `az login` and `logout`) and provide `--requirements` option that would install this plugin (from either git+http:// or local repo, as supported by pip), then passthrough the arguments to `gramine-sgx-sign` similarly to `$password` or so
2. if akv signing doesn't fit into signing plugins, then this needs to be rewritten that only parts of the template are changed, but the overall structure is not to be rewritten by the user

One possible way to do the second option would be to have `--template-dir` that would add an override `FileSystemLoader` like this:

```python
loader_gsc_templates = jinja2.FileSystemLoader('templates')
env = jinja2.Environment(loader=jinja2.ChoiceLoader([
    jinja2.PrefixLoader({'': loader_gsc_templates, }, '!'),
    jinja2.FileSystemLoader(user_provided_templates_path),
    loader_gsc_templates,
]))
```

Then you can instrument this signing templates with `{% block %}`s and inherit from the original templates:

```jinja
{% extends '!debian/sign.dockerfile' %}
{% block post-install %}
{{ super() }}
RUN python3 -m pip install whatever
RUN apt-get install also-whatever
{% endblock %}
```

```jinja
{% extends '!rhel/sign.dockerfile' %}
{% block post-install %}
{{ super() }}
RUN python3 -m pip install whatever
RUN yum install also-whatever-but-rpm
{% endblock %}
```

```jinja
{% extends '!sign.dockerfile' %}
{% block pre-sign %}
RUN az login
{% endblock %}

{% block post-sign %}
RUN az logout
{% endblock %}
```

But this would require reorganising the templates to allow not repeating ourselves, rethink names of templates and blocks, document those names, care about backwards compatibility etc.

PS 1b if support for encrypted keys is required, gramineproject/gramine#1197 is an RFC
</blockquote></details>

If that is your concern, let us go with my first proposal where we make the templates for already available HSMs, rather than in a separate contrib repo. At any point in time, there is going to be specific requirements be it AKV, or something else, and we will need to support it. Today we claim shipping a productization ready Gramine without having a production signing available. Let us solve that problem first for what we know, with what we already have as constraints.

And, how does your proposal of `--requirements` better than introducing other options which you initially opposed?

___

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/112)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @svenkata9)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/112#-NOxuuRf9tE_D_MPmcZ4:-NOzchYmAgzxKQbCdSsb:b-oyek60) (no related file):*
> we make the templates for already available HSMs

You can't possibly make templates for all available HSMs, because someone somewhere will do their own proprietary template for their own proprietary HSM. You've just created massive footgun for those people, if they don't really understand how docker works and how gsc works. This would be acceptable were it for some internal write-once software, but this is a public, FOSS project. We don't do footguns.

Ultimately, I don't see much technical difference between keeping auxiliary templates here or in contrib, they'll serve the same purpose for other people: they'll copy them and adjust to other workflows. Either way it's no good.

> And, how does your proposal of --requirements better than introducing other options which you initially opposed?

Because you don't allow random people anywhere near our dockerfiles and templates. This is something I desperately want to avoid. The option with inheritable templates tries to achieve the same thing, but it's a significantly bigger risk in term of APIs exposed that people will depend upon.

Yes, people will depend on it, and Hyrum's law is a real thing, that is, even if we don't promise anything about this interface, people will use it and will complain that we've broken it.

___

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/112)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @svenkata9)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/112#-NOxuuRf9tE_D_MPmcZ4:-NOzrL0H5hh5Q-Ch78M9:b4x275k) (no related file):*

> we make the templates for already available HSMs

You can't possibly make templates for all available HSMs, because someone somewhere will do their own proprietary template for their own proprietary HSM. You've just created massive footgun for those people, if they don't really understand how docker works and how gsc works. This would be acceptable were it for some internal write-once software, but this is a public, FOSS project. We don't do footguns.

Ultimately, I don't see much technical difference between keeping auxiliary templates here or in contrib, they'll serve the same purpose for other people: they'll copy them and adjust to other workflows. Either way it's no good.

> And, how does your proposal of --requirements better than introducing other options which you initially opposed?

Because you don't allow random people anywhere near our dockerfiles and templates. This is something I desperately want to avoid. The option with inheritable templates tries to achieve the same thing, but it's a significantly bigger risk in term of APIs exposed that people will depend upon.

Yes, people will depend on it, and Hyrum's law is a real thing, that is, even if we don't promise anything about this interface, people will use it and will complain that we've broken it.
</blockquote></details>

I've made a comment in gramineproject/contrib#20 at the exact line that's for me a NAK from security POV, but from maintainability all other exposed things like variable names are also problematic.

___

PR #113, opened 2022-12-14
Skip adding empty instruction `CMD [""]` and fix escaping

jkr0103:
## Description of the changes 

Change in this PR is to skip adding CMD instruction if there are no arguments to CMD in `Docker.build` file.
This fixes the issue as described in PR https://github.com/gramineproject/contrib/pull/21

I also tried the fix suggested by Woju in https://github.com/gramineproject/contrib/pull/21
It seems those filters are not available with jinja.

Another filter I tried is `tojson` like this: `{% if insecure_args and cmd %} CMD {{ cmd | tojson }} {% endif %}` but that doesn't give exactly same output as per original CMD:
CMD in base image: `CMD ["--protected-mode", "no", "--save", "''"]`
CMD in gsc/build/xxx/Dockerfile.build: `CMD ["--protected-mode", "no", "--save", "\u0027\u0027"]`

GSC output after building GSC without `--insecure-args`, i.e. putting args securely in `trusted_argsv` file:

*** FATAL CONFIG FILE ERROR (Redis 7.0.5) ***
Reading the configuration file, at line 3

'save "\u0027\u0027"'
Invalid save parameters

## How to test this PR? 
Create GSC image from a base image with/without CMD instructions and verity `gsc/build/image_name/Docker.build ` file. It should have CMD instructions if base image have CMD instruction with arguments.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)

### DIFF
--- a/gsc.py
+++ b/gsc.py
@@ -16,6 +16,7 @@ import tempfile
 
 import docker  # pylint: disable=import-error
 import jinja2
+import shlex
 import tomli   # pylint: disable=import-error
 import tomli_w # pylint: disable=import-error
 import yaml    # pylint: disable=import-error
@@ -76,9 +77,7 @@ def extract_binary_info_from_image_config(config, env):
     # Check if we have fixed binary arguments as part of entrypoint
     if num_starting_entrypoint_items > 1:
         last_bin_arg = num_starting_entrypoint_items
-        escaped_args = [s.replace('\\', '\\\\').replace('"', '\\"')
-                        for s in entrypoint[1:last_bin_arg]]
-        binary_arguments = '"' + '" "'.join(escaped_args) + '"'
+        binary_arguments = entrypoint[1:last_bin_arg]
     else:
         last_bin_arg = 0
         binary_arguments = ''
@@ -86,7 +85,6 @@ def extract_binary_info_from_image_config(config, env):
     # Place the remaining optional arguments previously specified as command in the new command.
     # Necessary since the first element of the command may be the binary of the resulting image.
     cmd = entrypoint[last_bin_arg + 1:] if len(entrypoint) > last_bin_arg + 1 else ''
-    cmd = [s.replace('\\', '\\\\').replace('"', '\\"') for s in cmd]
 
     env.globals.update({
         'binary': binary,
@@ -185,6 +183,7 @@ def gsc_build(args):
 
     # initialize Jinja env with configurations extracted from the original Docker image
     env = jinja2.Environment()
+    env.filters['shlex_quote'] = shlex.quote
     env.globals.update(config)
     env.globals.update(vars(args))
     env.globals.update({'app_image': original_image_name})
--- a/templates/Dockerfile.common.build.template
+++ b/templates/Dockerfile.common.build.template
@@ -44,7 +44,9 @@ COPY --chown={{app_user}} entrypoint.manifest /gramine/app_files/
 # Generate trusted arguments if required
 {% if not insecure_args %}
 RUN /gramine/meson_build_output/bin/gramine-argv-serializer \
-        /gramine/app_files/{{binary_basename}} {{binary_arguments}} "{{"\" \"".join(cmd)}}" > /gramine/app_files/trusted_argv
+        /gramine/app_files/{{binary_basename}} \
+        {{ binary_arguments | map('shlex_quote') | join(' ') }} \
+        {{ cmd | map('shlex_quote') | join(' ') }} > /gramine/app_files/trusted_argv
 {% endif %}
 
 # Docker entrypoint/cmd typically contains only the basename of the executable so create a symlink
@@ -61,4 +63,4 @@ RUN chmod u+x /gramine/app_files/apploader.sh \
 
 # Define default command
 ENTRYPOINT ["/bin/bash", "/gramine/app_files/apploader.sh"]
-CMD [{% if insecure_args %} "{{'", "'.join(cmd)}}" {% endif %}]
+{% if insecure_args and cmd %}CMD {{ cmd | tojson }}{% endif %}
--- a/templates/apploader.common.template
+++ b/templates/apploader.common.template
@@ -14,7 +14,9 @@ then
     gramine-sgx-get-token --quiet \
         --sig /gramine/app_files/entrypoint.sig --output /gramine/app_files/entrypoint.token
     gramine-sgx /gramine/app_files/entrypoint \
-        {% if insecure_args %}{{binary_arguments}} "${@}"{% endif %}
+        {% if insecure_args %}{{ binary_arguments | map('shlex_quote') | join(' ') }} \
+        "${@}"{% endif %}
 else
-    gramine-direct /gramine/app_files/entrypoint {{binary_arguments}} "${@}"
+    gramine-direct /gramine/app_files/entrypoint \
+    {{ binary_arguments | map('shlex_quote') | join(' ') }} "${@}"
 fi
--- a/test/ubuntu18.04-hello-world.dockerfile
+++ b/test/ubuntu18.04-hello-world.dockerfile
@@ -2,4 +2,4 @@ From ubuntu:18.04
 
 RUN apt-get update
 
-CMD ["echo", "\"Hello World!\""]
+CMD ["echo", "\"Hello World! Let's check escaped symbols: < & > \""]


woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jkr0103)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NJFvP2J-L9HSHaiFWTG:bgnvf6m) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*
This will break when there's `"` somewhere inside cmd.

Unless I'm missing something, it looks like the correct jinja filter is `cmd|map('tojson')|join(', ')`:

```
>>> jinja2.Template('CMD [{{ cmd|map("tojson")|join(", ") }}]').render(cmd=['echo', 'a', 'b', 'c d', 'e"f'])
'CMD ["echo", "a", "b", "c d", "e\\"f"]'
>>> jinja2.Template('CMD [{{ cmd|map("tojson")|join(", ") }}]').render(cmd=[])
'CMD []'
```

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @woju)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NJGLPT8ByBrd6l8p5Pf:bww6etb) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*

This will break when there's `"` somewhere inside cmd.

Unless I'm missing something, it looks like the correct jinja filter is `cmd|map('tojson')|join(', ')`:

```
>>> jinja2.Template('CMD [{{ cmd|map("tojson")|join(", ") }}]').render(cmd=['echo', 'a', 'b', 'c d', 'e"f'])
'CMD ["echo", "a", "b", "c d", "e\\"f"]'
>>> jinja2.Template('CMD [{{ cmd|map("tojson")|join(", ") }}]').render(cmd=[])
'CMD []'
```
</blockquote></details>

I tried the filter above with a image which have CMD [ "--protected-mode", "no", "--save ", "''" ]

and output is `CMD [ "--protected-mode", "no", "--save \u0027\u0027", "\u0027\u0027" ]` which is not exactly same. It produces similar results as the filter I mentioned in PR description.

And still even with new filter we get empty `CMD []`. Why do we need an empty `CMD []`? Changes here are to avoid adding empty CMD.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @woju)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NJGV2sfB5EzkEnp1_Wh:b-je6yy1) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*
>>> jinja2.Template('CMD [{{ cmd|map("tojson")|join(", ") }}]').render(cmd=['--protected-mode', 'no', '--save', '\'\''])
'CMD ["--protected-mode", "no", "--save", "\\u0027\\u0027"]'

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jkr0103)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NJGVlE23kGR5PT_EHp2:b-969kt6) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*

>>> jinja2.Template('CMD [{{ cmd|map("tojson")|join(", ") }}]').render(cmd=['--protected-mode', 'no', '--save', '\'\''])
'CMD ["--protected-mode", "no", "--save", "\\u0027\\u0027"]'

</blockquote></details>

Yes, looks correct to me. If you intentionally add `'` inside argument, they're meant to pass through. (`U+0027` is `APOSTROPHE` aka single quote).

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @woju)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NJGWzJ_4TS3V4CIQlVF:bbvrixj) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*

Yes, looks correct to me. If you intentionally add `'` inside argument, they're meant to pass through. (`U+0027` is `APOSTROPHE` aka single quote).
</blockquote></details>

Then I will use this filter.

Are you ok for avoiding addition of empty CMD instruction `CMD []`

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jkr0103)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NJGdwUj65Tozv7JsjoT:b-mm08pe) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*

Then I will use this filter.

Are you ok for avoiding addition of empty CMD instruction `CMD []`

</blockquote></details>

I don't know. If it doesn't break anything, it's probably OK.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NJGlXtI89c06ylnc5NH:bd03r1i) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*

I don't know. If it doesn't break anything, it's probably OK.
</blockquote></details>

I have added the filter and did some sanity tests, no failures observed.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NJKRxI03KGXtCmDJvQF:bib6suq) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*

I have added the filter and did some sanity tests, no failures observed.
</blockquote></details>

@woju Is there anything I need to add/modify? We have done multiple rounds of sanity on contrib as well as on GSC repo, no failures/issues observed.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NJLH7903kN4DMkuvy9d:b9qtirj) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*
@woju If we Build a redis image using below docker file and build directly with GSC:
`
Dockerfile:
FROM redis
CMD ["--protected-mode", "no", "--save", "''"]
`
Redis reports below error:
`
*** FATAL CONFIG FILE ERROR (Redis 7.0.5) ***
Reading the configuration file, at line 3
>>> 'save "\\u0027\\u0027"'
Invalid save parameters
`
But run successfully without this PR changes.

This particular filter doesn't handle `APOSTROPHE` properly, it seems but older filter doesn't have issue.

On your comment `This will break when there's " somewhere inside cmd.`
Actually this is handled in `gsc.py` starting from here: 
https://github.com/gramineproject/gsc/blob/dbc17b6a118d0fb13411d3acd1f1dfc032b77b83/gsc.py#L79

So it seems old filter works fine except adding `""` if cmd is empty which we can handle using `if else` case as in first commit of this PR.
What do you say, revert to old filter and use first commit in this PR?

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NJOmMyP3fBj0p3JPoyZ:b-3qvgg9) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*

@woju If we Build a redis image using below docker file and build directly with GSC:
`
Dockerfile:
FROM redis
CMD ["--protected-mode", "no", "--save", "''"]
`
Redis reports below error:
`
*** FATAL CONFIG FILE ERROR (Redis 7.0.5) ***
Reading the configuration file, at line 3
>>> 'save "\\u0027\\u0027"'
Invalid save parameters
`
But run successfully without this PR changes.

This particular filter doesn't handle `APOSTROPHE` properly, it seems but older filter doesn't have issue.

On your comment `This will break when there's " somewhere inside cmd.`
Actually this is handled in `gsc.py` starting from here: 
https://github.com/gramineproject/gsc/blob/dbc17b6a118d0fb13411d3acd1f1dfc032b77b83/gsc.py#L79

So it seems old filter works fine except adding `""` if cmd is empty which we can handle using `if else` case as in first commit of this PR.
What do you say, revert to old filter and use first commit in this PR?

</blockquote></details>

https://jinja.palletsprojects.com/en/2.10.x/templates/#tojson
as per above link`tojson` escape the following characters in strings:
`<, >, &, '` which is safe for `<script>` tags.

I tried all these characters in CMD as 
`CMD ["echo", "\"Hello World! Less_than:<, Greater_than:>, Bitwise_and:&, Apostrophe:' \""]`
and got output: `"Hello World! Less_than:\u003c, Greater_than:\u003e, Bitwise_and:\u0026, Apostrophe:\u0027 "`

but will first commit in this PR, it works as expected.

I seems tojson is design to be compatible with html tags. If we get these characters in CMD, `tojson` filter won't be suitable here.

dimakuv (changes requested):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NKryyjN35oIbOQFmXze:b560yfn) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*

https://jinja.palletsprojects.com/en/2.10.x/templates/#tojson
as per above link`tojson` escape the following characters in strings:
`<, >, &, '` which is safe for `<script>` tags.

I tried all these characters in CMD as 
`CMD ["echo", "\"Hello World! Less_than:<, Greater_than:>, Bitwise_and:&, Apostrophe:' \""]`
and got output: `"Hello World! Less_than:\u003c, Greater_than:\u003e, Bitwise_and:\u0026, Apostrophe:\u0027 "`

but will first commit in this PR, it works as expected.

I seems tojson is design to be compatible with html tags. If we get these characters in CMD, `tojson` filter won't be suitable here.
</blockquote></details>

From my side, I would go with the original commit in this PR -- the one that simply added a `if cmd` check to avoid outputting an empty `CMD []`.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NKvaUCHEiHvxZwku_C5:bkcj385) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*

From my side, I would go with the original commit in this PR -- the one that simply added a `if cmd` check to avoid outputting an empty `CMD []`.
</blockquote></details>

Reverted to original commit.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @woju)

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @jkr0103)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NKx5YQ36aEhCFOxN6bQ:b-auwd38) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*

Reverted to original commit.
</blockquote></details>

No, this is in any case incorrect. Try with `"` somewhere inside cmd or `\` at the end of some argument and see what happens.

If `tojson` doesn't work, then you probably need to use correct escaping function, possibly https://docs.python.org/3/library/shlex.html#shlex.quote or `repr()`? (I don't know the exact quoting rules for dockerfiles to authoritatively say which functions is OK). If the correct functions is not available in template languate, then it needs to be added to env's globals or functions.

https://jinja.palletsprojects.com/en/3.1.x/api/#global-namespace
https://jinja.palletsprojects.com/en/3.1.x/api/#writing-filters
https://jinja.palletsprojects.com/en/3.1.x/api/#jinja2.Environment.globals
https://jinja.palletsprojects.com/en/3.1.x/api/#jinja2.Environment.filters

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @jkr0103)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NKxTiWc3pWKCqLTpley:bkj7nd3) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*
I did some testing and I think `tojson()` is the correct escaping function.

For one, Docker reference says about `CMD` (https://docs.docker.com/engine/reference/builder/#cmd):

> The exec form is parsed as a JSON array, which means that you must use double-quotes (`"`) around words not single-quotes (`'`).

(exec form is `CMD ["command", "arg1", "arg2"]`)

And sure enough, it works as intended, even with `\u` escapes:

```
% cat Dockerfile.jinja 
FROM debian:latest

CMD ["echo", {{ 'ĄĆĘŁŃÓŚŹŻ'|tojson() }}]
% cat render.py 
#!/usr/bin/python3

import jinja2

env = jinja2.Environment(loader=jinja2.FileSystemLoader('.'))

with open('Dockerfile', 'w') as file:
    file.write(env.get_template('Dockerfile.jinja').render())
% cat Dockerfile
FROM debian:latest

CMD ["echo", "\u0104\u0106\u0118\u0141\u0143\u00d3\u015a\u0179\u017b"]
% docker run --rm -it fbe2df9d9692
ĄĆĘŁŃÓŚŹŻ
```

IDK what's wrong with it, I suppose there's too much escaping somewhere else.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @woju)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NKxW_aZCO5FLAJSqeIq:b-67rj3c) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*
> No, this is in any case incorrect. Try with " somewhere inside cmd or \ at the end of some argument and see what happens.
@woju this is taken care here https://github.com/gramineproject/gsc/blob/dbc17b6a118d0fb13411d3acd1f1dfc032b77b83/gsc.py#L89
and here https://github.com/gramineproject/gsc/blob/dbc17b6a118d0fb13411d3acd1f1dfc032b77b83/gsc.py#L79

and these scenarios are already tested with helloworld test example in gsc, here https://github.com/gramineproject/gsc/blob/dbc17b6a118d0fb13411d3acd1f1dfc032b77b83/test/ubuntu18.04-hello-world.dockerfile#L5

>IDK what's wrong with it, I suppose there's too much escaping somewhere else.
extra escaping come from the above two lines in gsc.py.

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @jkr0103)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NKxYYfACVlt8RtLjR6M:b-ynvila) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*
> this is taken care here 

Oh. No, this is very wrong. Escaping should be done in a single place, not "this character we take care of here, and this issue there".

Escaping is a function of end format (different formats have different rules for escaping), so it should be done in jinja in template, using functions exposed from python if needed. Doing some half-escapes here and there is a recipe for disaster.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: 0 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NL1T5tr-XfTw3QWT3oT:b-81aeq8) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*

Yep. I agree with @woju, and after looking into it a bit more, here is my proposal:

- We remove all in-Python ad-hoc escaping for `binary_arguments` and `cmd`:
    - https://github.com/gramineproject/gsc/blob/dbc17b6a118d0fb13411d3acd1f1dfc032b77b83/gsc.py#L79-L81
    - https://github.com/gramineproject/gsc/blob/dbc17b6a118d0fb13411d3acd1f1dfc032b77b83/gsc.py#L89

- We unify `binary_arguments` to be a Python list, similar to how `cmd` is now:
    - in other words, we remove this string concatenation: https://github.com/gramineproject/gsc/blob/dbc17b6a118d0fb13411d3acd1f1dfc032b77b83/gsc.py#L81

- We use Woju's initial proposal of `{{ cmd|map("tojson")|join(", ") }}` -- this works like charm. It escapes all required symbols and it adds double-quotes around each item.

- We will need to fix the following lines in Dockerfile templates:
    - https://github.com/gramineproject/gsc/blob/dbc17b6a118d0fb13411d3acd1f1dfc032b77b83/templates/Dockerfile.common.build.template#L47
    - https://github.com/gramineproject/gsc/blob/dbc17b6a118d0fb13411d3acd1f1dfc032b77b83/templates/Dockerfile.common.build.template#L64
    - https://github.com/gramineproject/gsc/blob/dbc17b6a118d0fb13411d3acd1f1dfc032b77b83/templates/apploader.common.template#L17
    - https://github.com/gramineproject/gsc/blob/dbc17b6a118d0fb13411d3acd1f1dfc032b77b83/templates/apploader.common.template#L19
</blockquote></details>

Implemented the changes as suggested and tested with bash helloworld, python and redis.

dimakuv (dismissed):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @woju)

woju (changes requested):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @jkr0103)

___
*[`-- commits` line 2 at r5](https://reviewable.io/reviews/gramineproject/gsc/113#-NLLPDEVADasYFV4p7uW):*
Commit message is outdated, because it now fixes escaping, not only skips empty CMD.
___
*[`templates/Dockerfile.common.build.template` line 48 at r1](https://reviewable.io/reviews/gramineproject/gsc/113#-NJFvP2I8xr4FMr8T6b7:-NLLOzAi44_F8XOkuXBj:b-85i63v) ([raw file](https://github.com/gramineproject/gsc/blob/4f8ee440957c67a54bbd63346b9e9cedb5e92f68/templates/Dockerfile.common.build.template#L48)):*

Implemented the changes as suggested and tested with bash helloworld, python and redis.
</blockquote></details>

LGTM.

dimakuv (approved):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/113)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @woju)

___
*[`-- commits` line 2 at r5](https://reviewable.io/reviews/gramineproject/gsc/113#-NLLPDEVADasYFV4p7uW):*

Commit message is outdated, because it now fixes escaping, not only skips empty CMD.
</blockquote></details>

Done
___
*[`test/ubuntu18.04-hello-world.dockerfile` line 5 at r6](https://reviewable.io/reviews/gramineproject/gsc/113#-NLLoCyM44jgDi0UQ_YV:-NLLoCyM44jgDi0UQ_YW:b-3xxnax) ([raw file](https://github.com/gramineproject/gsc/blob/fd19073b063ddb2b329ea849eba3c098d1887a2f/test/ubuntu18.04-hello-world.dockerfile#L5)):*
> ```Dockerfile
> RUN apt-get update
> 
> CMD ["echo", "\"Hello World! Let's check escaped symbols: < & > \""]
> ```

FYI: This is just for quick testing of escaping (both with `./gsc build --insecure-args` and with `./gsc build`).

I tested manually, it works with this PR.

PR #120, opened 2023-01-20
Add symlink files to the trusted files list

### LINKED ISSUE #25: [finalize_manifest.py] Finalize manifest - traverse symlink directories
Converting the official mongo docker image to gsc fails because of the symlink /lib -> usr/lib, the finalize_manifest.py doesn't include the /lib directory in the manifest, due to os.walk not traversing symlinks, but later on the application tries to access the /lib directory.

### LINKED ISSUE #116: Graminizing the Ubuntu 20.04 image fails with /bin/bash: Permission denied
## Description of the problem
Creating a graminized image of Ubuntu 20.04 with bash fails with /bin/bash: Permission denied. 

## Steps to reproduce

1. Create a test.sh with echo command
2. Create a Dockerfile with Entrypoint/CMD as /bin/bash test.sh and base os as 20.04
3. Build the docker base image
4. Graminize the application

**test.sh**
```
#!/bin/sh

echo "Hello Bash Shell Script"
/bin/bash sample.sh
```

**sample.sh**
```
#!/bin/sh

echo "Hello Sample Shell Script"
```

**Dockerfile:**
```
From ubuntu:20.04

RUN apt-get update

COPY test.sh .
COPY sample.sh .

CMD ["/bin/bash", "test.sh"]
```

## Expected results
When graminized application is running, it should return **"Hello Bash Shell Script"**

## Actual results
docker run  --device=/dev/sgx/enclave -it gsc-bash-test
```
Gramine is starting. Parsing TOML manifest file, this may take some time...
/usr/local/bin/entry_script_bash.sh: line 4: /bin/bash: Permission denied
```

## Additional information
In all ubuntu:20.04 based docker images, /bin is symlink to /usr/bin. Because of this GSC doesn't add any /bin file to trusted files.  Any bash script executed at run time having shebang #!/bin/bash is not able to access /bin/bash as it's not part of trusted files list. We get error:  /bin/bash: bad interpreter: Permission denied".

jkr0103:
GSC doesn't add any symlink file to trusted list. In all the ubuntu:20.04 based docker images, `/bin` is symlink to `/usr/bin` and because of this GSC doesn't add any `/bin` file to trusted files.

Any bash script executed at run time having shebang `#!/bin/bash` is not able to access `/bin/bash` as it's not part of trusted files list, reports error: `/bin/bash: bad interpreter: Permission denied`.

As part of the fix, GSC will add all symlink files.

Fixes #116, fixes #25

I tested Curated-Apps (Redis, PyTorch, Sklearn and Tensorflow-serving) with this PR using gramine v1.3.1 and v1.4.

All the Curated-Apps on gramine v1.4 works without requiring any change in `enclave_size` and `pal_internal_memory`.

On gramine 1.3.1, Pytorch needed extra 64M `pal_internal_memory` in release mode and it did not work in debug mode even after increasing `pal_internal_memory` upto 2G. This seems to an issue with gramine 1.3.1 which doesn't show up in gramine latest master. Please refer below table for more details.

| Workload  | Build    |Trusted File, no symlinks 1.3.1/1.4|Trusted File with symlinks 1.3.1/1.4|% increment 1.3.1/1.4|`pal_internal_memory` in gramine 1.3.1 / 1.4|
| ------------| -------  | -------------------------:  | ----------------------------: | -------------:  | ------------------------------------|
| Redis         | release | 19115 / 19194                | 20950 / 21029                   |  10% / 10%    | no change / no change               |
|                   | debug  | 92587 / 104188              | 94430 / 106031                 |    2% /  2%    | no change / no change                |
| PyTorch     | release | 99330 / 99409                 | 165798 / 165877               |  67% / 67%   | +64M / no change                       |
|                   | debug  | 171027 / 182628            |  237504 / 249105              |  39% / 36%    |  gramine hang / no change         |
| Scikit-learn| release | 41061 / 41140                |  43188 / 43267                  |   5% / 5%       | no change / no change                |
|                   | debug  | 115052 / 126653            | 115052 / 128788               |   2% / 2%       | no change / no change                |
| TF-serving | release | 20248 / 20327                | 23076 / 23155                   |  23% / 23%     | no change / no change               |

## How to test this PR? 

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)

### DIFF
--- a/finalize_manifest.py
+++ b/finalize_manifest.py
@@ -52,9 +52,23 @@ def generate_trusted_files(root_dir, already_added_files):
 
     num_trusted = 0
     trusted_files = []
-    for root, _, files in os.walk(root_dir.encode('UTF-8'), followlinks=False):
+
+    # WARNING: below loop does not remove cycles (which can happen due to symlinks), but we assume
+    # that Docker images (excluding paths/files listed in `excluded_paths_regex`) don't have cycles.
+    for dirpath, dirnames, files in os.walk(root_dir.encode('UTF-8'), followlinks=True):
+        scandirs = []
+        for dirname in dirnames:
+            to_be_walked_dir = os.path.join(dirpath, dirname)
+            if exclude_re.match(to_be_walked_dir.decode('UTF-8')):
+                # exclude special paths from list of trusted files
+                continue
+            scandirs.append(dirname)
+        # slice assignment, to modify `dirnames` in place (so that os.walk() will recurse only
+        # into the subdirs whose names remain in `dirnames`)
+        dirnames[:] = scandirs
+
         for file in files:
-            filename = os.path.join(root, file)
+            filename = os.path.join(dirpath, file)
             if not os.path.isfile(filename):
                 # only regular files are added as trusted files
                 continue


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jkr0103)

___
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/gramineproject/gsc/120#-NMT9c4P10-EeZ1foOQq):*
I would also add an explanation as the commit body message:
```
Since `os.walk(..., followlinks=True)` can lead to infinite recursion
due to symlinks pointing to each other, we need to keep track of
already-visited directories via the `dirs` set.
```
___
*[`finalize_manifest.py` line 64 at r1](https://reviewable.io/reviews/gramineproject/gsc/120#-NMT935R1ZKkGmzSfV9V:-NMT935R1ZKkGmzSfV9W:b-m9sy64) ([raw file](https://github.com/gramineproject/gsc/blob/2bbc3b4b3dbb8711fea300994618cf558978c2ca/finalize_manifest.py#L64)):*
> ```Python
>                 continue
>             st = os.stat(dir)
>             dirkey = st.st_dev, st.st_ino, dir
> ```

I know we discussed this offline, but I still don't understand why we need to use `dev` and `ino` as additional identifiers of "is it the path we already traversed?"...

What happens if we remove `st_dev` and `st_ino`? Wouldn't just `dir` (full path name) be enough?

Then the code will also look simpler and nicer:
```
        ...
        for dirname in dirnames:
            dir = os.path.join(dirpath, dirname)
            if exclude_re.match(dir.decode('UTF-8')):
                # exclude special paths from list of trusted files
                continue
            if dir not in dirs:
                dirs.add(dir)
                scandirs.append(dirname)
        dirnames[:] = scandirs
```

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/gramineproject/gsc/120#-NMT9c4P10-EeZ1foOQq):*

I would also add an explanation as the commit body message:
```
Since `os.walk(..., followlinks=True)` can lead to infinite recursion
due to symlinks pointing to each other, we need to keep track of
already-visited directories via the `dirs` set.
```
</blockquote></details>

Done.
___
*[`finalize_manifest.py` line 64 at r1](https://reviewable.io/reviews/gramineproject/gsc/120#-NMT935R1ZKkGmzSfV9V:-NMUEdsuEbvIQwsH0mXo:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/2bbc3b4b3dbb8711fea300994618cf558978c2ca/finalize_manifest.py#L64)):*

I know we discussed this offline, but I still don't understand why we need to use `dev` and `ino` as additional identifiers of "is it the path we already traversed?"...

What happens if we remove `st_dev` and `st_ino`? Wouldn't just `dir` (full path name) be enough?

Then the code will also look simpler and nicer:
```
        ...
        for dirname in dirnames:
            dir = os.path.join(dirpath, dirname)
            if exclude_re.match(dir.decode('UTF-8')):
                # exclude special paths from list of trusted files
                continue
            if dir not in dirs:
                dirs.add(dir)
                scandirs.append(dirname)
        dirnames[:] = scandirs
```
</blockquote></details>

Done.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/120#-NMXw2cEBj2ZpjyPN9d3:-NMXw2cF5-Y_NWWwjqeb:b-2wgfxz) (no related file):*
@jkr0103 I know you experimented with this PR a bit on different workloads. Could you share your results? Like, how many more files are added to `sgx.trusted_files` with this PR, and whether it requires greater `enclave_size` in Gramine manifest?

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/120#-NMXw2cEBj2ZpjyPN9d3:-NMc5qGO5t8YZJgNdoxb:bct07vq) (no related file):*

I have attached the report in PR description.
</blockquote></details>

Thanks, looks good to me.

___

mkow (changes requested):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @jkr0103)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbhIlfD4Lh3OsoP_ha:-NNbhIlfD4Lh3OsoP_hb:b-gkyv5n) (no related file):*
@jkr0103: Please don't paste screenshots / attach excel files if you just want to include a simple table. Use Markdown for this.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbix5BERUdcMN6F5aN:-NNbix5BERUdcMN6F5aO:b-i8l4cp) (no related file):*
Do I understand correctly, that this PR fixes a few workloads, but breaks PyTorch? If that's right, then please debug and fix it.

___
___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NNbkIqfB8YP87b8UUXF:b-dmlh00) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*
> ```Python
>                 # exclude special paths from list of trusted files
>                 continue
>             if dir not in dirs:
> ```

@woju: Could you verify if this logic is correct for traversing directory tree with symlinks?
___
*[`finalize_manifest.py` line 66 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbjWkc7gnnhwYHab7A:-NNbjWkc7gnnhwYHab7B:blq1v7i) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L66)):*
> ```Python
>                 dirs.add(dir)
>                 scandirs.append(dirname)
>         dirnames[:] = scandirs
> ```

How does this work?



_Code quote:_
```Python
dirnames[:] = scandirs
```

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbhIlfD4Lh3OsoP_ha:-NNepPsQ827idIakkIHH:bb7f2mf) (no related file):*

@jkr0103: Please don't paste screenshots / attach excel files if you just want to include a simple table. Use Markdown for this.
</blockquote></details>

done

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbix5BERUdcMN6F5aN:-NNepTuj-gmDENZ7C4i4:b-spf9lq) (no related file):*
> If that's right, then please debug and fix it.
Pytorch workload, only in debug breaks on gramine 1.3.1 but works on gramine master master. Do we really need to root cause that in gramine 1.3.1 which is not seen in gramine latest master, may be as we don't have to add `pal_internal_memory` as it's depricated. Hence I don't see the issue is blocker here.

___
___
*[`finalize_manifest.py` line 66 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbjWkc7gnnhwYHab7A:-NNeqYOF0Be5j7z7JSf-:by3fs89) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L66)):*
it just replaces the list with new one, example below:

```
>>> p = ['P','y','t','h','o','n']
>>> p
['P', 'y', 't', 'h', 'o', 'n']
>>> p[:] = [1, 2, 3]
>>> p
[1, 2, 3]
```

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @mkow)

___
*[`finalize_manifest.py` line 59 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNs5AFmBtclPYWYorLw:-NNs5AFmBtclPYWYorLx:b-vkgl4y) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L59)):*
> ```Python
>         scandirs = []
>         for dirname in dirnames:
>             dir = os.path.join(dirpath, dirname)
> ```

`dir()` is a builtin (even reviewable highlighted it), please rename this variable
___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NNs7hOuEZAWjGzK_24P:b-lnj6rr) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

@woju: Could you verify if this logic is correct for traversing directory tree with symlinks?
</blockquote></details>

Depending on what you're attempting to do? Do you want to deduplicate them or copy? `os.walk` is naive, will traverse a directory twice if there are two links pointing to it (or one subdirectory and a symlink pointing to it). `dirs = set()` and this `if` will not prevent it, because the path is what `os.walk` went through, not the `realpath`. This `if` is noop (always true) if that's what you're asking, and the logic of replacing `dirnames` looks pointless.

But it goes worse, `os.walk` will happily traverse link loops and will turn around the DFS when one `scandir` throws `OSError` for `ELOOP` (40), in which case `os.walk` will catch this exception and silently (!) ignore it and just won't follow further.

Try it yourself:
```
mkdir -p a/b
ln -s ../.. a/b/c
python3 -c "for d, _, _ in __import__('os').walk('a', followlinks=True): print(d)"
```

In the case from PR description (`/bin/` in usrmerged system) it will probably have 2 copies of trusted files (same hashes for 2 different paths), but it's very hard to say if there aren't any unintended consequences in other FS tree layouts. In absolute sense, this is not a "correct logic" from POSIX pov, but I don't know enough about handling symlinks in Gramine and GSC to say if this should be rejected altogether. This `if` is certainly to be either removed, or corrected with `os.path.realpath()` or `pathlib.Path.resolve()`.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NNu_eSV7DSjksiyd47h:b-n839o5) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*
> In the case from PR description (`/bin/` in usrmerged system) it will probably have 2 copies of trusted files (same hashes for 2 different paths)

@woju This is exactly what we want to achieve.

> it's very hard to say if there aren't any unintended consequences in other FS tree layouts

I'm unsure what you mean by this? From the Gramine perspective, whatever happens on the host is irrelevant, as long as Gramine can access the host file using the path.

> `os.walk` is naive, will traverse a directory twice if there are two links pointing to it (or one subdirectory and a symlink pointing to it).

Yes, and this is exactly what we want here. We want to traverse the same directory twice, if there are two links (two "path names") leading to it.

> This `if` is noop (always true) if that's what you're asking

I did some quick tests, Woju is right.

So looks like the `if` should be like this:
```python
        # instead of `if dir not in dirs:`
        st = os.stat(dir)
        dirkey = st.st_dev, st.st_ino
        if dirkey not in dirs:
            dirs.add(dirkey)
            scandirs.append(dirname)
```

@woju Does this check make more sense to you?

---

Basically, the whole point of this PR is to this scenario:

- `/bin/bash` is an executable, and `/bin/` is the "real" directory.
- `/usr/bin/` is a symlink to `/bin/`; so `/usr/bin/bash` is the same executable as `/bin/bash`.

Gramine needs to add both absolute paths to `sgx.trusted_files` in the manifest:
```
sgx.trusted_files = [
  "file:/bin/bash",
  "file:/usr/bin/bash",
]
```

This is literally the only thing we try to achieve with this PR. We don't care that both paths lead to the same file. We don't care that a prefix of one path is a symlink to the other path. We rely on the host to follow symlinks. (And Gramine treats these two as just normal files, Gramine has no idea/concept of symlinks.)

dimakuv:
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NNugUVW9eskBww-m5WK:brw6yxt) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

> In the case from PR description (`/bin/` in usrmerged system) it will probably have 2 copies of trusted files (same hashes for 2 different paths)

@woju This is exactly what we want to achieve.

> it's very hard to say if there aren't any unintended consequences in other FS tree layouts

I'm unsure what you mean by this? From the Gramine perspective, whatever happens on the host is irrelevant, as long as Gramine can access the host file using the path.

> `os.walk` is naive, will traverse a directory twice if there are two links pointing to it (or one subdirectory and a symlink pointing to it).

Yes, and this is exactly what we want here. We want to traverse the same directory twice, if there are two links (two "path names") leading to it.

> This `if` is noop (always true) if that's what you're asking

I did some quick tests, Woju is right.

So looks like the `if` should be like this:
```python
        # instead of `if dir not in dirs:`
        st = os.stat(dir)
        dirkey = st.st_dev, st.st_ino
        if dirkey not in dirs:
            dirs.add(dirkey)
            scandirs.append(dirname)
```

@woju Does this check make more sense to you?

---

Basically, the whole point of this PR is to this scenario:

- `/bin/bash` is an executable, and `/bin/` is the "real" directory.
- `/usr/bin/` is a symlink to `/bin/`; so `/usr/bin/bash` is the same executable as `/bin/bash`.

Gramine needs to add both absolute paths to `sgx.trusted_files` in the manifest:
```
sgx.trusted_files = [
  "file:/bin/bash",
  "file:/usr/bin/bash",
]
```

This is literally the only thing we try to achieve with this PR. We don't care that both paths lead to the same file. We don't care that a prefix of one path is a symlink to the other path. We rely on the host to follow symlinks. (And Gramine treats these two as just normal files, Gramine has no idea/concept of symlinks.)
</blockquote></details>

Well, my proposal doesn't work too.

Here is my quick test, that contains both the scenario we want to achieve (`bash` accessible through two paths) and an infinite link:
```sh
mkdir -p usr/bin
touch usr/bin/bash
ln -s usr/bin bin
ln -s ../.. usr/bin/inflink
```

This gives us the following tree:
```sh
$ tree
.
├── bin -> usr/bin
└── usr
    └── bin
        ├── bash
        └── inflink -> ../..
```

This is the Python snippet that I used for testing, resembling closely what GSC wants to do:
```python
import os
dirs = set()
for dirpath, dirnames, files in os.walk('.', followlinks=True):
    scandirs = []
    for dirname in dirnames:
        dir = os.path.join(dirpath, dirname)
        st = os.stat(dir)
        dirkey = st.st_dev, st.st_ino
        if dirkey not in dirs:
            dirs.add(dirkey)
            scandirs.append(dirname)
    dirnames[:] = scandirs
    for file in files:
        filename = os.path.join(dirpath, file)
        if not os.path.isfile(filename):
            # only regular files are added as trusted files
            continue
        print(filename)
```

This is the result of running this script:
```sh
./bin/bash

# there is no ./usr/bin/bash !!!
```

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NNuouybE5kgChwRghlb:bf474zp) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

Well, my proposal doesn't work too.

Here is my quick test, that contains both the scenario we want to achieve (`bash` accessible through two paths) and an infinite link:
```sh
mkdir -p usr/bin
touch usr/bin/bash
ln -s usr/bin bin
ln -s ../.. usr/bin/inflink
```

This gives us the following tree:
```sh
$ tree
.
├── bin -> usr/bin
└── usr
    └── bin
        ├── bash
        └── inflink -> ../..
```

This is the Python snippet that I used for testing, resembling closely what GSC wants to do:
```python
import os
dirs = set()
for dirpath, dirnames, files in os.walk('.', followlinks=True):
    scandirs = []
    for dirname in dirnames:
        dir = os.path.join(dirpath, dirname)
        st = os.stat(dir)
        dirkey = st.st_dev, st.st_ino
        if dirkey not in dirs:
            dirs.add(dirkey)
            scandirs.append(dirname)
    dirnames[:] = scandirs
    for file in files:
        filename = os.path.join(dirpath, file)
        if not os.path.isfile(filename):
            # only regular files are added as trusted files
            continue
        print(filename)
```

This is the result of running this script:
```sh
./bin/bash

# there is no ./usr/bin/bash !!!
```
</blockquote></details>

This is a really hard problem :( Or I'm that stupid.

The only hacky solution that I could come up is to allow X repetitions of the same path (same inode). For example, in the code below I assume 2 repetitions:
```python
import os
dirs = set()
for dirpath, dirnames, files in os.walk('.', followlinks=True):
    scandirs = []
    for dirname in dirnames:
        dir = os.path.join(dirpath, dirname)
        st = os.stat(dir)
        dirkey1 = st.st_dev, st.st_ino, 1  # first repetition
        dirkey2 = st.st_dev, st.st_ino, 2  # second repetition
        if dirkey1 not in dirs:
            dirs.add(dirkey1)
            scandirs.append(dirname)
        elif dirkey2 not in dirs:
            dirs.add(dirkey2)
            scandirs.append(dirname)
    dirnames[:] = scandirs
    for file in files:
        filename = os.path.join(dirpath, file)
        if not os.path.isfile(filename):
            # only regular files are added as trusted files
            continue
        print(filename)
```

This gives me correct output to the example I posted earlier:
```sh
./usr/bin/bash
./bin/bash
```

But of course if we'd have one more symlink (e.g. `ln -s usr/bin one-more-bin`), then my script would fail to register the abs path `one-more-bin/bash`.

@mkow @woju Any smart ideas?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @jkr0103, and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NNur62u8XJH_AGmvCen:b-w6kkbc) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

This is a really hard problem :( Or I'm that stupid.

The only hacky solution that I could come up is to allow X repetitions of the same path (same inode). For example, in the code below I assume 2 repetitions:
```python
import os
dirs = set()
for dirpath, dirnames, files in os.walk('.', followlinks=True):
    scandirs = []
    for dirname in dirnames:
        dir = os.path.join(dirpath, dirname)
        st = os.stat(dir)
        dirkey1 = st.st_dev, st.st_ino, 1  # first repetition
        dirkey2 = st.st_dev, st.st_ino, 2  # second repetition
        if dirkey1 not in dirs:
            dirs.add(dirkey1)
            scandirs.append(dirname)
        elif dirkey2 not in dirs:
            dirs.add(dirkey2)
            scandirs.append(dirname)
    dirnames[:] = scandirs
    for file in files:
        filename = os.path.join(dirpath, file)
        if not os.path.isfile(filename):
            # only regular files are added as trusted files
            continue
        print(filename)
```

This gives me correct output to the example I posted earlier:
```sh
./usr/bin/bash
./bin/bash
```

But of course if we'd have one more symlink (e.g. `ln -s usr/bin one-more-bin`), then my script would fail to register the abs path `one-more-bin/bash`.

@mkow @woju Any smart ideas?
</blockquote></details>

If you're after expansion of all the symlinks, then just drop the resolving logic? Let the walk, like, walk all the files incl. symlinks? Or am I missing something?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NNuxiNg1tuOiQR58lSZ:b-pba9fg) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

If you're after expansion of all the symlinks, then just drop the resolving logic? Let the walk, like, walk all the files incl. symlinks? Or am I missing something?
</blockquote></details>

@woju But you're suggesting basically your snippet that goes bananas on loop-links? This one:
```
python3 -c "for d, _, _ in __import__('os').walk('a', followlinks=True): print(d)"
```

This approach is probably working (I think the current state of this PR boils down to this?), but then we'll add ~40x more files than necessary. Sounds like something nobody wants to discover in their final `sgx.trusted_files`.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @jkr0103, and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NNv073MANUcZ0Ke00Y5:bdk901a) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

@woju But you're suggesting basically your snippet that goes bananas on loop-links? This one:
```
python3 -c "for d, _, _ in __import__('os').walk('a', followlinks=True): print(d)"
```

This approach is probably working (I think the current state of this PR boils down to this?), but then we'll add ~40x more files than necessary. Sounds like something nobody wants to discover in their final `sgx.trusted_files`.
</blockquote></details>

Either I'm gravely misunderstanding something, or this is the only correct approach. If the expectation is that you can access the underlying file by any of those paths, they should be included, right?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NNv1Gp43TSbwerG1doh:b-9ewfqt) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*
> If the expectation is that you can access the underlying file by any of those paths, they should be included, right?

If we're talking about the purist approach, then you're right, this is the only correct approach.

However, no sane application should open files through these nested symlinks (`a/b/c/a/b/c/a/b/c/a`). So ideally, I'd like an approach that **removes back edges** (removes cycles in the graph of files). Unfortunately, at least some popular Docker images (iirc, PyTorch) have such nested symlinks.

But writing an algo in Python for detecting cycles in traversed files graph... sounds like a total overkill.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @jkr0103, and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NNv7tGc6VMrhzlBUI00:b3udvk2) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

> If the expectation is that you can access the underlying file by any of those paths, they should be included, right?

If we're talking about the purist approach, then you're right, this is the only correct approach.

However, no sane application should open files through these nested symlinks (`a/b/c/a/b/c/a/b/c/a`). So ideally, I'd like an approach that **removes back edges** (removes cycles in the graph of files). Unfortunately, at least some popular Docker images (iirc, PyTorch) have such nested symlinks.

But writing an algo in Python for detecting cycles in traversed files graph... sounds like a total overkill.
</blockquote></details>

I think yes, we are talking about purist approaches and POSIX correctness. The GSC is a general tool which should work with any docker doing whatever it likes, isn't it?

I think if you want to optimize this 40x, then we should think about adding symlink support to trusted files. We'd add another type of trusted file that would not only expose a symlink inside, but would also affect the resolution, i.e. we'd have some equivalent of `realpath(3)` in trusted PAL that would resolve the path to a canonical version and verify the checksum from that.

Either way, I think GSC is a wrong place to fix this.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NNvCuO79YlVuLB-DNnv:b-xsgakb) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*
> I think if you want to optimize this 40x, then we should think about adding symlink support to trusted files. We'd add another type of trusted file that would not only expose a symlink inside, but would also affect the resolution, i.e. we'd have some equivalent of realpath(3) in trusted PAL that would resolve the path to a canonical version and verify the checksum from that.

Yes, we've been thinking about this. But adding links in Gramine is a non-trivial task, requiring a lot of thought and implementation effort.

But I agree in general that GSC should just work with arbitrary Docker images, without any specific hacks.

@jkr0103 So actually what happens if we simply flip `followlinks=False` to `followlinks=True` in normal GSC (without this PR)? Why won't it just work (though it may take *much more time* and have *much bigger manifest file size*)?

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NO80s9p0GAwf5uKJvaW:bwp2jm6) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

> I think if you want to optimize this 40x, then we should think about adding symlink support to trusted files. We'd add another type of trusted file that would not only expose a symlink inside, but would also affect the resolution, i.e. we'd have some equivalent of realpath(3) in trusted PAL that would resolve the path to a canonical version and verify the checksum from that.

Yes, we've been thinking about this. But adding links in Gramine is a non-trivial task, requiring a lot of thought and implementation effort.

But I agree in general that GSC should just work with arbitrary Docker images, without any specific hacks.

@jkr0103 So actually what happens if we simply flip `followlinks=False` to `followlinks=True` in normal GSC (without this PR)? Why won't it just work (though it may take *much more time* and have *much bigger manifest file size*)?
</blockquote></details>

That was the first thing I tried, and `os.walk` keep running and files processed were `125,705,000` after 1 hour and still not over. Saw multiple loops in in `/sys` and one of them was 
 `/sys/kernel/software_nodes/node0/dmi-ipmi-si.0`. With the current changes, we exclude `/sys` etc., hence we don't see that happening. I tested this is bash example.

Ideal approach wouold be to `removes cycles in the graph of files`. Shell I try that?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NO95JQp-BshA1X3QZDI:bwo3ryh) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*
> Ideal approach wouold be to removes cycles in the graph of files. Shell I try that?

Can you try it in another PR?

I think at this point we should do the following:
1. Create a new PR that simply flips `followlinks=True` and test it extensively (just like you did with this PR)
2. Create a new PR with item 1 plus "remove-the-cycles" logic (but honestly, this feels like a total overkill)
3. Mark this PR as blocked, we'll most probably close it in favor of solution 1 or 2

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @jkr0103, and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NO9d5ou3zgCRbsQLV_p:bj6z1th) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

> Ideal approach wouold be to removes cycles in the graph of files. Shell I try that?

Can you try it in another PR?

I think at this point we should do the following:
1. Create a new PR that simply flips `followlinks=True` and test it extensively (just like you did with this PR)
2. Create a new PR with item 1 plus "remove-the-cycles" logic (but honestly, this feels like a total overkill)
3. Mark this PR as blocked, we'll most probably close it in favor of solution 1 or 2
</blockquote></details>

I think you should start with a list of directories that you certainly don't want measured, like `/proc`, `/sys`, `/tmp`, `/run`, `/var/run` etc. See if it solves the problem.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NO9ec6754yXbuPMFNIF:b-ijv5d) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

I think you should start with a list of directories that you certainly don't want measured, like `/proc`, `/sys`, `/tmp`, `/run`, `/var/run` etc. See if it solves the problem.
</blockquote></details>

Current approach in PR exclude the directories which we don't want to measure, but it doesn't avoid loops (cycles). Do we want to fix that as part of this PR?

Another change left is renaming `dir` keyword.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NO9iTFPDF7szpjDf_6_:b-uv4d1h) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

Current approach in PR exclude the directories which we don't want to measure, but it doesn't avoid loops (cycles). Do we want to fix that as part of this PR?

Another change left is renaming `dir` keyword.
</blockquote></details>

I agree with @woju. If the simple approach works, then let's just use it.

Avoiding loops (a) requires non-trivial Python code and a lot of testing, and (b) is significantly slower than the simple approach.

That's why I suggested to create a new PR with the logic that avoids loops (if you really want to have this functionality), so we can have two approaches and can compare them in terms of number of trusted files and performance.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: 0 of 1 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NOAa75h0J1vmUK3652n:bi16dmo) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

I agree with @woju. If the simple approach works, then let's just use it.

Avoiding loops (a) requires non-trivial Python code and a lot of testing, and (b) is significantly slower than the simple approach.

That's why I suggested to create a new PR with the logic that avoids loops (if you really want to have this functionality), so we can have two approaches and can compare them in terms of number of trusted files and performance.
</blockquote></details>

Code here https://github.com/gramineproject/gsc/blob/7d94c5857903ede2a8a9b14d6211c354f28f173b/finalize_manifest.py#L69 is for excluding the files but loop exists at directory level even before this code executes and we get stuck there. Hence we still need to loop over `dirnames` and exclude from this list those dirs that matches.

Only case it doesn't fix is infinite loop created using below commands, hope such a loop doesn't appear in real-world workloads.



_Code snippet:_
```Python
mkdir -p usr/bin
touch usr/bin/bash
ln -s usr/bin bin
ln -s ../.. usr/bin/inflink
```

dimakuv (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NOEEDyp6L2qRIcwsQyZ:bd5s0zz) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

Code here https://github.com/gramineproject/gsc/blob/7d94c5857903ede2a8a9b14d6211c354f28f173b/finalize_manifest.py#L69 is for excluding the files but loop exists at directory level even before this code executes and we get stuck there. Hence we still need to loop over `dirnames` and exclude from this list those dirs that matches.

Only case it doesn't fix is infinite loop created using below commands, hope such a loop doesn't appear in real-world workloads.


</blockquote></details>

Yes, I agree with @jkr0103, we still need to loop over the sub-directories' list and filter out the "not needed ones". The current solution looks good to me.

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @mkow)

___
*[`finalize_manifest.py` line 59 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNs5AFmBtclPYWYorLw:-NOAj9o58CvwUZils0JM:bhcdxp) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L59)):*

done
</blockquote></details>

Resolving, but this is not really `abs_`: the path will be either relative or absolute, depending on what was `root_dir`. Or did you intend this to be absolute? Then you need to explicitly call 
___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NOAig151yjPDncHq4m2:bph9car) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

Yes, I agree with @jkr0103, we still need to loop over the sub-directories' list and filter out the "not needed ones". The current solution looks good to me.
</blockquote></details>

If we're going to settle on this solution, I'd like a comment in the source about this behaviour.

Also, this won't really be an infinite loop, it will end with `ELOOP`, that is, when path length hits some arbitrary limit (around 256 bytes), kernel will error out any operation on such path, so no risk it will be stuck forever.
___
*[`finalize_manifest.py` line 50 at r3](https://reviewable.io/reviews/gramineproject/gsc/120#-NOB-kKw8LDpIEsz49aE:-NOB-kKw8LDpIEsz49aF:bbwg3rw) ([raw file](https://github.com/gramineproject/gsc/blob/4b5b9cf8ebc7993d20e6cbfaf686cd458772bd72/finalize_manifest.py#L50)):*
> ```Python
>                                 r'|proc/.*'
>                                 r'|sys/.*'
>                                 r'|var/.*)$')
> ```

WRT this list:
1) `/var/.*` is wrong, what about `/var/lib`, `/var/db` etc.?
2) `/run` (and `/var/run`, one of them is probably a symlink to the other) is missing.

And this is just from the top of my head, it needs to be checked with FHS and some variety of real distros, as different as possible.

woju:
*[`finalize_manifest.py` line 59 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNs5AFmBtclPYWYorLw:-NOEikur3n0H31Az3bMo:bf30xxj) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L59)):*

Resolving, but this is not really `abs_`: the path will be either relative or absolute, depending on what was `root_dir`. Or did you intend this to be absolute? Then you need to explicitly call 
</blockquote></details>

... call `os.path.abspath()` or `pathlib.Path.resolve` or show that `root_dir` is always absolute.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[`finalize_manifest.py` line 59 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNs5AFmBtclPYWYorLw:-NOEzB6S2_ncUoHYiqWQ:bjmhehw) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L59)):*

... call `os.path.abspath()` or `pathlib.Path.resolve` or show that `root_dir` is always absolute.
</blockquote></details>

No, this was not supposed to be absolute. You're right, I'd rename to `to_be_walked_dir`.
___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NOEznf46YlZpuwo8NhD:b-kp57da) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

If we're going to settle on this solution, I'd like a comment in the source about this behaviour.

Also, this won't really be an infinite loop, it will end with `ELOOP`, that is, when path length hits some arbitrary limit (around 256 bytes), kernel will error out any operation on such path, so no risk it will be stuck forever.
</blockquote></details>

@jkr0103 Could you add a comment right-before the line with `os.walk()` like this:
```
# WARNING: below loop does not remove cycles (which can happen due to symlinks), but we assume
# that Docker images (excluding paths/files listed in `excluded_paths_regex`) don't have cycles.
```
___
*[`finalize_manifest.py` line 50 at r3](https://reviewable.io/reviews/gramineproject/gsc/120#-NOB-kKw8LDpIEsz49aE:-NOF-rBU-trh1gSUcS4T:b-kxgdf5) ([raw file](https://github.com/gramineproject/gsc/blob/4b5b9cf8ebc7993d20e6cbfaf686cd458772bd72/finalize_manifest.py#L50)):*

WRT this list:
1) `/var/.*` is wrong, what about `/var/lib`, `/var/db` etc.?
2) `/run` (and `/var/run`, one of them is probably a symlink to the other) is missing.

And this is just from the top of my head, it needs to be checked with FHS and some variety of real distros, as different as possible.
</blockquote></details>

We should create a separate PR to improve this list. @woju Any ideas what would be good references/links to populate this list?

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: 0 of 1 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @woju)

___
*[`finalize_manifest.py` line 59 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNs5AFmBtclPYWYorLw:-NOF2_dG2oTQ7i_v7_uH:brythjh) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L59)):*

No, this was not supposed to be absolute. You're right, I'd rename to `to_be_walked_dir`.
</blockquote></details>

renamed.
___
*[`finalize_manifest.py` line 63 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbkIqfB8YP87b8UUXE:-NOF2dLEANfh1PbjiM1O:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L63)):*

@jkr0103 Could you add a comment right-before the line with `os.walk()` like this:
```
# WARNING: below loop does not remove cycles (which can happen due to symlinks), but we assume
# that Docker images (excluding paths/files listed in `excluded_paths_regex`) don't have cycles.
```
</blockquote></details>

Done.

dimakuv (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

woju (dismissed):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`finalize_manifest.py` line 50 at r3](https://reviewable.io/reviews/gramineproject/gsc/120#-NOB-kKw8LDpIEsz49aE:-NOFwRX06hC0Z3gCZTdG:bdh937x) ([raw file](https://github.com/gramineproject/gsc/blob/4b5b9cf8ebc7993d20e6cbfaf686cd458772bd72/finalize_manifest.py#L50)):*

We should create a separate PR to improve this list. @woju Any ideas what would be good references/links to populate this list?
</blockquote></details>

#128 with references for you (or reassign if you want).

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`finalize_manifest.py` line 50 at r3](https://reviewable.io/reviews/gramineproject/gsc/120#-NOB-kKw8LDpIEsz49aE:-NOJKss92K2kHF29TDf0:b-zcjvte) ([raw file](https://github.com/gramineproject/gsc/blob/4b5b9cf8ebc7993d20e6cbfaf686cd458772bd72/finalize_manifest.py#L50)):*

#128 with references for you (or reassign if you want).
</blockquote></details>

Thanks

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @jkr0103, @mkow, and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbix5BERUdcMN6F5aN:-NOKcIw35uqMZlZKIveK:b162vi8) (no related file):*

> We didn't root cause the exact reason, because this problem disappeared in the latest master branch.

So, it works on v1.4?
</blockquote></details>

Yes, it works on v1.4. Actually, now that we have the release `v1.4`, would be good to put this version in the `config.yaml` of GSC and try with this (instead of `master`).

___
___
*[`-- commits` line 7 at r4](https://reviewable.io/reviews/gramineproject/gsc/120#-NOKf20kEx0iHImhB7f1):*
We need to modify the commit body message:
```
This is done via `os.walk(..., followlinks=True)`. This does not remove cycles
(which can happen due to symlinks), but we assume that Docker images
(excluding paths/files listed in `excluded_paths_regex`) don't have cycles.
```
___
*[`finalize_manifest.py` line 66 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbjWkc7gnnhwYHab7A:-NOKgZld5Tx6UIJU3xv5:b-7986bz) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L66)):*

Ouch, this is really underhanded, I don't know any other standard Python APIs working this way...
Maybe pathlib have something more sane?
</blockquote></details>

Maybe @woju knows?
___
*[`finalize_manifest.py` line 57 at r4](https://reviewable.io/reviews/gramineproject/gsc/120#-NOKafkt3k8-nSGml3U1:-NOKgVhu0H7N1svzcNt3:b-3p9p1d) ([raw file](https://github.com/gramineproject/gsc/blob/e096246869f6300a213c0ffc0e837da9c7dc18a2/finalize_manifest.py#L57)):*

This comment doesn't match the commit message? Which one is true?
</blockquote></details>

See my other comment, I proposed the new commit message.

jkr0103 (commented):
Tested the PR with gramine v1.4 in `config.yaml` with new implementation and updated the table.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120#-:-NOL6h_z6AHcXWPdjZ3B:b5hdk24)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbix5BERUdcMN6F5aN:-NOL6IXi6_gKgFRLPhDk:blld0kv) (no related file):*

Yes, it works on v1.4. Actually, now that we have the release `v1.4`, would be good to put this version in the `config.yaml` of GSC and try with this (instead of `master`).
</blockquote></details>

I used gramine v1.4 in `config.yaml` and tested the Curated-Apps workloads in release and debug modes and have updated the table. All works fine.

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @jkr0103, @mkow, and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbix5BERUdcMN6F5aN:-NOLADs9-YFhKSjeyLQm:b-ucfknh) (no related file):*

I used gramine v1.4 in `config.yaml` and tested the Curated-Apps workloads in release and debug modes and have updated the table. All works fine.
</blockquote></details>

Thanks!

___

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @mkow)

___
*[`finalize_manifest.py` line 66 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbjWkc7gnnhwYHab7A:-NOOrs1z5-xLmv9NW0k5:b-gxrfa) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L66)):*

Maybe @woju knows?
</blockquote></details>

This is not insane, there are many places where you're supposed to modify an object that you're handed from within stdlib. The difference here is that you're handed a `list`, not some custom class instance to modify attributes.

`pathlib` doesn't have anything better. There's https://docs.python.org/3/library/pathlib.html#pathlib.Path.rglob, but you can't tell it to not descend into one subtree like we do here.

If you're after readability and you dislike slice assignment, you can do it like this:

```python
for i in range(len(dirnames), -1, -1):
    if that_big_regex.fullmatch(os.path.join(dirpath, dirnames[i])):
        dirnames.pop(i)
```

But beware: if you instead do it like this, it's unsound:

```python
for dirname in dirnames:
    if that_big_regex.fullmatch(os.path.join(dirpath, dirname)):
        dirnames.remove(dirname)
```

Because after first `.remove()` iterator is invalid (might have glitched). So in the future you need to be careful with reviews, to make sure that no-one ever makes it "more pythonic".

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 4 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @mkow)

___
*[`finalize_manifest.py` line 66 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbjWkc7gnnhwYHab7A:-NOP-Jvu4dM_V-91XYL_:bxx0jd2) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L66)):*

This is not insane, there are many places where you're supposed to modify an object that you're handed from within stdlib. The difference here is that you're handed a `list`, not some custom class instance to modify attributes.

`pathlib` doesn't have anything better. There's https://docs.python.org/3/library/pathlib.html#pathlib.Path.rglob, but you can't tell it to not descend into one subtree like we do here.

If you're after readability and you dislike slice assignment, you can do it like this:

```python
for i in range(len(dirnames), -1, -1):
    if that_big_regex.fullmatch(os.path.join(dirpath, dirnames[i])):
        dirnames.pop(i)
```

But beware: if you instead do it like this, it's unsound:

```python
for dirname in dirnames:
    if that_big_regex.fullmatch(os.path.join(dirpath, dirname)):
        dirnames.remove(dirname)
```

Because after first `.remove()` iterator is invalid (might have glitched). So in the future you need to be careful with reviews, to make sure that no-one ever makes it "more pythonic".
</blockquote></details>

Based on what @woju explained, I would keep the current code as-is. We may add a quick comment to this line, like:
```
# slice assignment, to modify the list `dirnames` in place
```

WDYT @mkow ?

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @jkr0103, and @woju)

___
*[`finalize_manifest.py` line 66 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbjWkc7gnnhwYHab7A:-NOPXAGB2lAgvee9-_jb:bgrj2k9) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L66)):*
> The difference here is that you're handed a `list`, not some custom class instance to modify attributes.

Yes, that's what I mean. My problem is not about slice assignment, but about library returning a list which you then modify during iteration to influence the future iterations.

A comment like Dmitrii suggest would be good, but should rather say that this changes how `os.walk()` will continue iteration.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @woju)

___
*[`finalize_manifest.py` line 66 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbjWkc7gnnhwYHab7A:-NOQ5tXx3eGSrP1CELC_:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L66)):*

> The difference here is that you're handed a `list`, not some custom class instance to modify attributes.

Yes, that's what I mean. My problem is not about slice assignment, but about library returning a list which you then modify during iteration to influence the future iterations.

A comment like Dmitrii suggest would be good, but should rather say that this changes how `os.walk()` will continue iteration.
</blockquote></details>

Done.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @mkow)

___
*[`finalize_manifest.py` line 66 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbjWkc7gnnhwYHab7A:-NOQavZMC9lMf6s1_G7e:bo1q9ub) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L66)):*

Done.
</blockquote></details>

The comment looks good to me.

woju (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 7 at r4](https://reviewable.io/reviews/gramineproject/gsc/120#-NOKf20kEx0iHImhB7f1):*

We need to modify the commit body message:
```
This is done via `os.walk(..., followlinks=True)`. This does not remove cycles
(which can happen due to symlinks), but we assume that Docker images
(excluding paths/files listed in `excluded_paths_regex`) don't have cycles.
```
</blockquote></details>

I concur.
___
*[`finalize_manifest.py` line 66 at r2](https://reviewable.io/reviews/gramineproject/gsc/120#-NNbjWkc7gnnhwYHab7A:-NOVjB-V3i8kyR-xmm9j:b-fml7ds) ([raw file](https://github.com/gramineproject/gsc/blob/06842315825dbaa03c7305a1f4eac2f237327734/finalize_manifest.py#L66)):*

The comment looks good to me.
</blockquote></details>

yeah, LGTM

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/120#-NOiq2qu5UyRaIyKQc74:-NOiq2qu5UyRaIyKQc75:b-pnl45b) (no related file):*
I did a final rebase and tested that it works correctly.

___
___
*[`-- commits` line 7 at r4](https://reviewable.io/reviews/gramineproject/gsc/120#-NOKf20kEx0iHImhB7f1):*

I concur.
</blockquote></details>

Done

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/120)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #130, opened 2023-02-16
DONTMERGE Fix `excluded_paths_regex` in `finalize_manifest.py`

### LINKED ISSUE #128: Fix `excluded_paths_regex` in `finalize_manifest.py`
This is split from #120.

At least some of `/var` needs to be preserved, `/run` needs to be discarded. Very likely other paths.

References:

- https://manpages.debian.org/testing/manpages/hier.7.en.html
- https://manpages.debian.org/testing/systemd/file-hierarchy.7.en.html
- https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html
- https://www.debian.org/doc/debian-policy/ch-opersys.html#file-system-hierarchy
- https://docs.fedoraproject.org/en-US/packaging-guidelines/#_filesystem_layout

dimakuv:
## Description of the changes 

The updated list of excluded "trusted files" paths better follows the Filesystem Hierarchy Standard (FHS).

Based on:
- https://manpages.debian.org/testing/manpages/hier.7.en.html
- https://manpages.debian.org/testing/systemd/file-hierarchy.7.en.html
- https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html

Fixes #128.

Additionally, this PR includes a quick second commit, which sets `USER` in `ubuntu18.04-hello-world` test. This is for quick tests of non-root images.

## How to test this PR? 

Test some Docker images, no functional changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gsc/130)

### DIFF
--- a/Documentation/index.rst
+++ b/Documentation/index.rst
@@ -257,24 +257,22 @@ follows three main stages and produces an image named ``gsc-<image-name>``.
 #. **Graminizing the application image.** The second stage copies the important
    Gramine artifacts (e.g., the runtime and signer tool) from the first stage
    (or if the first stage was skipped, it pulls a prebuilt Docker image defined
-   via the configuration file).  It then prepares image-specific variables such
+   via the configuration file). It then prepares image-specific variables such
    as the executable path and the library path, and scans the entire image to
-   generate a list of trusted files.  GSC excludes files and paths starting with
-   :file:`/boot`, :file:`/dev`, :file:`.dockerenv`, :file:`.dockerinit`,
-   :file:`/etc/mtab`, :file:`/etc/rc`, :file:`/proc`, :file:`/sys`, and
-   :file:`/var`, since checksums are required which either don't exist or may
-   vary across different deployment machines. GSC combines these variables and
-   list of trusted files into a new manifest file. In a last step the entrypoint
-   is changed to launch the :file:`apploader.sh` script which generates an Intel
-   SGX token (only if needed, on non-FLC platforms) and starts the
-   :program:`gramine-sgx` loader. Note that the generated image
-   (``gsc-<image-name>-unsigned``) cannot successfully load an Intel SGX
-   enclave, since essential files and the signature of the enclave are still
-   missing (see next stage).
+   generate a list of trusted files. GSC excludes some system files and paths
+   (for the exact list, see :ref:`excluded-paths`), since checksums are required
+   which either don't exist or may vary across different deployment machines.
+   GSC combines these variables and list of trusted files into a new manifest
+   file. In a last step the entrypoint is changed to launch the
+   :file:`apploader.sh` script which generates an Intel SGX token (only if
+   needed, on non-FLC platforms) and starts the :program:`gramine-sgx` loader.
+   Note that the generated image (``gsc-<image-name>-unsigned``) cannot
+   successfully load an Intel SGX enclave, since essential files and the
+   signature of the enclave are still missing (see next stage).
 
 #. **Signing the Intel SGX enclave.** The third stage uses Gramine's signer
    tool to generate SIGSTRUCT files for SGX enclave initialization. This tool
-   also generates an SGX-specific manifest file.  The required signing key is
+   also generates an SGX-specific manifest file. The required signing key is
    provided by the user via the :command:`gsc sign-image` command and copied
    into this Docker build stage. The generated image is called
    ``gsc-<image-name>`` and includes all necessary files to start an Intel SGX
@@ -518,14 +516,35 @@ either in environment variables or mounted as files. GSC is currently unaware of
 such files and hence, cannot mark them trusted. Similar to trusted data, these
 files may be added to the manifest.
 
+.. _excluded-paths:
+
 Access to files in excluded paths
 ---------------------------------
 
-The manifest generation excludes all files and paths starting with :file:`/boot`
-, :file:`/dev`, :file:`.dockerenv`, :file:`.dockerinit`, :file:`/etc/mtab`,
-:file:`/etc/rc`, :file:`/proc`, :file:`/sys`, and :file:`/var` from the list of
-trusted files. If your application relies on some files in these directories,
-you must manually add them to the manifest::
+The manifest generation excludes the following files and paths from the list of
+trusted files:
+
+- :file:`/boot/*`
+- :file:`/dev/*`
+- :file:`/efi/*`
+- :file:`/media/*`
+- :file:`/mnt/*`
+- :file:`/proc/*`
+- :file:`/run/*`
+- :file:`/sys/*`
+- :file:`/tmp/*`
+- :file:`/var/*`
+- security-critical files under :file:`/etc/`:
+
+  - :file:`/etc/.pwd.lock`
+  - :file:`/etc/gshadow`
+  - :file:`/etc/mtab`
+  - :file:`/etc/rc*`
+  - :file:`/etc/security`
+  - :file:`/etc/shadow`
+
+If your application relies on some files in these directories, you must manually
+add them to the manifest::
 
    sgx.trusted_files = [ "file:file1", "file:file2" ]
    or
--- a/finalize_manifest.py
+++ b/finalize_manifest.py
@@ -32,21 +32,27 @@ def extract_files_from_user_manifest(manifest):
 
 
 def generate_trusted_files(root_dir, already_added_files):
+    # please keep this list in sync with the one in GSC documentation
     excluded_paths_regex = (r'^/('
-                                r'boot/.*'
-                                r'|\.dockerenv'
-                                r'|\.dockerinit'
+                                r'|boot/.*'
                                 r'|dev/.*'
+                                r'|efi/.*'
+                                # below files are security-critical (only root can access them), and
+                                # trying to hash them in non-root Docker images would fail
+                                r'|etc/\.pwd\.lock'
                                 r'|etc/gshadow.*'
                                 r'|etc/mtab'
-                                r'|etc/\.pwd\.lock'
                                 r'|etc/rc(\d|.)\.d/.*'
                                 r'|etc/security/.*'
                                 r'|etc/shadow.*'
-                                r'|gramine/python/.*'
+                                # below file will be removed in final GSC Docker image, so skip it
                                 r'|gramine/app_files/finalize_manifest\.py'
+                                r'|media/.*'
+                                r'|mnt/.*'
                                 r'|proc/.*'
+                                r'|run/.*'
                                 r'|sys/.*'
+                                r'|tmp/.*'
                                 r'|var/.*)$')
     exclude_re = re.compile(excluded_paths_regex)
 
--- a/test/ubuntu18.04-hello-world.dockerfile
+++ b/test/ubuntu18.04-hello-world.dockerfile
@@ -1,5 +1,7 @@
 From ubuntu:18.04
 
-RUN apt-get update
+RUN apt-get -y update
+RUN groupadd -r user && useradd -r -g user user
+USER user
 
 CMD ["echo", "\"Hello World! Let's check escaped symbols: < & > \""]


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/130)* status: 0 of 3 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`finalize_manifest.py` line 56 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NOPkap4-KJBdAxqlXVZ:-NOPkap4-KJBdAxqlXV_:bq8wsj4) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/finalize_manifest.py#L56)):*
> ```Python
>                                 r'|sys/.*'
>                                 r'|tmp/.*'
>                                 r'|var/.*)$')
> ```

@woju thinks we need to exclude only some subdirs of `/var/` (or better, include only some subdirs of `/var/`). But I didn't find any useful subdirs for "normal" apps, typically converted by GSC.

dimakuv:
@jkr0103 @anjalirai-intel @aneessahib Does this PR make sense to you?

dimakuv (commented):
@aneessahib 

> An additional thought - would it make sense to give the flexibility to the user to add more exclude paths via a proper template? This will help users create smaller GSC images as well.

Sounds like a bit too much... Unless you have a simple idea how users will provide such excluded paths in command-line args (or in some other way)?

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/130#-:-NOQb6Le5hUytpN6MjgC:bdrrmak)* status: 0 of 3 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

aneessahib (commented):
via separate file template. like a file `exclude_path.template` for example ? We are having to tell multiple people to manually edit `finalize_manifest.py` which is deep in GSC. A documented interface might help. Anyways, thats a separate thing to address. 

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/130#-:-NOTiRLX7DB0pnwRfWxp:b-21nbfj)* status: 0 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`Documentation/index.rst` line 264 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NOTjw9cCAkciGYHXu27:-NOTjw9cCAkciGYHXu28:b21tnd6) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/Documentation/index.rst#L264)):*
> ```reStructuredText
>    generate a list of trusted files. GSC excludes some system files and paths
>    (for the exact list, see :ref:`excluded-paths`), since checksums are required
>    which either don't exist or may vary across different deployment machines.
> ```

I think somewhere here we should point to `finalize_manifest.py`. So that users can modify the exception list according to their needs?

dimakuv (commented):
> We are having to tell multiple people to manually edit `finalize_manifest.py` which is deep in GSC.

But why? Why do they have to modify it? What's breaking?

> via separate file template. like a file `exclude_path.template` for example?

I'm not sure how this should work. Could you give an example of such a template? Would it be a list of regexes?

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/130#-:-NOUJMzo5d57ZQUYbBhO:bx7j2sw)* status: 0 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @aneessahib)

___
*[`Documentation/index.rst` line 264 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NOTjw9cCAkciGYHXu27:-NOUJabhAp1sNqhQo8np:bfhh8h4) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/Documentation/index.rst#L264)):*

I think somewhere here we should point to `finalize_manifest.py`. So that users can modify the exception list according to their needs?
</blockquote></details>

No, `finalize_manifest.py` is an internal detail of GSC. We shouldn't tell users to modify the GSC source code.

woju (changes requested):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/130)* status: 1 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @aneessahib and @dimakuv)

___
*[`finalize_manifest.py` line 36 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NOklObf5-vmbq_ep314:-NOklObf5-vmbq_ep315:b-5y1m5l) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/finalize_manifest.py#L36)):*
> ```Python
> def generate_trusted_files(root_dir, already_added_files):
>     # please keep this list in sync with the one in GSC documentation
>     excluded_paths_regex = (r'^/('
> ```

While you're at it, can you reformat it?

- drop hanging indent, it's wasting too much horizontal space and makes inline comments difficult
- rewrite as multiline string for `re.VERBOSE` (https://docs.python.org/3/library/re.html#re.VERBOSE)
- drop common `/` before `(` (it's confusing)
- drop `^` at the beginning and `.*` at the end (`.match()` method takes care of those)
- inline `re.compile`:

```python
exclude_re = re.compile(r'''
    ( /boot
    | /dev
    | /efi

    # below files are security-critical
    | /etc/\.pwd\.lock

    | /gramine/app_files/finalize_manifest\.py # will be removed in final GSC image
    )
''', re.VERBOSE)

exclude_re.match(...) # this is already OK below
```

Or alternatively you can rework this as a list of `fnmatch` aka "glob" patterns (https://docs.python.org/3/library/fnmatch.html), which might be better if you want to add CLI-level configuration (`--include-files "/var/lib/*"`, `--exclude-files /etc/\*passwd`), because many CLI tools have fnmatch-like patterns here, not regexps (like https://manpages.debian.org/bullseye/tar/tar.1.en.html#exclude or https://manpages.debian.org/bullseye-backports/rsync/rsync.1.en.html#PATTERN_MATCHING_RULES).

```python
exclude_globs = [
    '/boot',
    '/dev',
    '/media/*',
    #...
]

if any(fnmatch.fnmatch(to_be_walked_dir, pattern) for pattern in exclude_globs):
    ...
```

___
*[`finalize_manifest.py` line 56 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NOPkap4-KJBdAxqlXVZ:-NOkjFH37TbnyhWih5e4:blfydpd) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/finalize_manifest.py#L56)):*

@woju thinks we need to exclude only some subdirs of `/var/` (or better, include only some subdirs of `/var/`). But I didn't find any useful subdirs for "normal" apps, typically converted by GSC.
</blockquote></details>

From the top of my head:
- Most HTTP servers: `/var/www`
- Postgresql: `/var/lib/postgresql` (I guess no-one runs it, because shm, but it's a data point)
- OpenLDAP: `/var/lib/ldap`
- ...

Now some of them may not be appropriate (e.g. postfix uses `/var/lib/postfix` to store lockfiles, when it should've used `/run`). But certainly there are some things that are in fact needed. Esp. when they're the data that is actually served (`/var/www`). But if you think about adding some configurability, you'd have to add *two* mechanisms, one for adding to exclude_re, and second for "ouch, I needed this directory after all".
___
*[`Documentation/index.rst` line 264 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NOTjw9cCAkciGYHXu27:-NOkoGhP1cw4rNgT1Kqb:b-9sygww) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/Documentation/index.rst#L264)):*
> We shouldn't tell users to modify the GSC source code.

Yes.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/130)* status: 1 of 3 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @aneessahib and @dimakuv)

___
*[`finalize_manifest.py` line 56 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NOPkap4-KJBdAxqlXVZ:-NOmh8lp6QMKJUOIJ3CT:b4u88pl) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/finalize_manifest.py#L56)):*
> @woju thinks we need to exclude only some subdirs of /var/ (or better, include only some subdirs of /var/). But I didn't find any useful subdirs for "normal" apps, typically converted by GSC.

I don't think there is a right list of "userful subdirs" for "normal" apps. And if we exclude only some subdirs of `/var/`, then what about `/run/` (isn't it a similar case: see https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html)?

> But if you think about adding some configurability, you'd have to add two mechanisms

I think it's useful but not sure whether we do this in this PR (or as a follow-up).
___
*[`test/ubuntu18.04-hello-world.dockerfile` line 4 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NOmcMtA07zPeiUahOMc:-NOmcMtBCbp2A4s5flxK:b-xna8dz) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/test/ubuntu18.04-hello-world.dockerfile#L4)):*
> ```Dockerfile
> 
> RUN apt-get -y update
> RUN groupadd -r user && useradd -r -g user user
> ```

Any specific considerations that we apply `-r` here?



_Code quote:_
```Dockerfile
RUN groupadd -r user && useradd -r -g user user
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/130)* status: 1 of 3 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @aneessahib, @kailun-qin, and @woju)

___
*[`finalize_manifest.py` line 36 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NOklObf5-vmbq_ep314:-NOnChDrA2He4DiyWqz-:b-v9p27r) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/finalize_manifest.py#L36)):*

While you're at it, can you reformat it?

- drop hanging indent, it's wasting too much horizontal space and makes inline comments difficult
- rewrite as multiline string for `re.VERBOSE` (https://docs.python.org/3/library/re.html#re.VERBOSE)
- drop common `/` before `(` (it's confusing)
- drop `^` at the beginning and `.*` at the end (`.match()` method takes care of those)
- inline `re.compile`:

```python
exclude_re = re.compile(r'''
    ( /boot
    | /dev
    | /efi

    # below files are security-critical
    | /etc/\.pwd\.lock

    | /gramine/app_files/finalize_manifest\.py # will be removed in final GSC image
    )
''', re.VERBOSE)

exclude_re.match(...) # this is already OK below
```

Or alternatively you can rework this as a list of `fnmatch` aka "glob" patterns (https://docs.python.org/3/library/fnmatch.html), which might be better if you want to add CLI-level configuration (`--include-files "/var/lib/*"`, `--exclude-files /etc/\*passwd`), because many CLI tools have fnmatch-like patterns here, not regexps (like https://manpages.debian.org/bullseye/tar/tar.1.en.html#exclude or https://manpages.debian.org/bullseye-backports/rsync/rsync.1.en.html#PATTERN_MATCHING_RULES).

```python
exclude_globs = [
    '/boot',
    '/dev',
    '/media/*',
    #...
]

if any(fnmatch.fnmatch(to_be_walked_dir, pattern) for pattern in exclude_globs):
    ...
```

</blockquote></details>

Ok, I really like @woju's proposal on `fnmatch` and the new `--include-files`/`--exclude-files` command-line options. I think I should close this PR, as the proposal is *significantly* different from what this PR tries to achieve.

Someone (maybe me...) will need to implement and test Woju's proposal.
___
*[`finalize_manifest.py` line 56 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NOPkap4-KJBdAxqlXVZ:-NOnCzF67KoUzCDFIF9r:b-9vdisa) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/finalize_manifest.py#L56)):*
<details><summary><i>Previously, kailun-qin (Kailun Qin) wrote…</i></summary><blockquote>

> @woju thinks we need to exclude only some subdirs of /var/ (or better, include only some subdirs of /var/). But I didn't find any useful subdirs for "normal" apps, typically converted by GSC.

I don't think there is a right list of "userful subdirs" for "normal" apps. And if we exclude only some subdirs of `/var/`, then what about `/run/` (isn't it a similar case: see https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html)?

> But if you think about adding some configurability, you'd have to add two mechanisms

I think it's useful but not sure whether we do this in this PR (or as a follow-up).
</blockquote></details>

Yeah, ok, you're both right. This all needs much better handling. See my other comment.
___
*[`Documentation/index.rst` line 264 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NOTjw9cCAkciGYHXu27:-NOnD4uR45eLLd8whAUA:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/Documentation/index.rst#L264)):*

> We shouldn't tell users to modify the GSC source code.

Yes.
</blockquote></details>

Done.
___
*[`test/ubuntu18.04-hello-world.dockerfile` line 4 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NOmcMtA07zPeiUahOMc:-NOnCV-z4tiw_ze2GMGx:b6ynn3c) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/test/ubuntu18.04-hello-world.dockerfile#L4)):*
<details><summary><i>Previously, kailun-qin (Kailun Qin) wrote…</i></summary><blockquote>

Any specific considerations that we apply `-r` here?
</blockquote></details>

I took this example from Docker documentation, but I agree, I don't see any reason to create a "system" user. I could remove it.

woju (commented):
> Reviewed 2 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/130)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @aneessahib and @kailun-qin)

mkow (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/130)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @aneessahib, @dimakuv, @kailun-qin, and @woju)

___
*[`test/ubuntu18.04-hello-world.dockerfile` line 5 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NPcHIAd9Fy-g0MyDPL5:-NPcHIAd9Fy-g0MyDPL6:bq0lppd) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/test/ubuntu18.04-hello-world.dockerfile#L5)):*
> This is for quick tests of non-root images.

Won't this break images which assume root? I.e. aren't you fixing some images and breaking some others at the same time?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/130)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @aneessahib, @kailun-qin, and @woju)

___
*[`test/ubuntu18.04-hello-world.dockerfile` line 5 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NPcHIAd9Fy-g0MyDPL5:-NPqMMqu7KGSKqg8mSDZ:b-1adn99) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/test/ubuntu18.04-hello-world.dockerfile#L5)):*

> This is for quick tests of non-root images.

Won't this break images which assume root? I.e. aren't you fixing some images and breaking some others at the same time?
</blockquote></details>

Sorry @mkow I don't understand the question. Why did you pose the question here, in the HelloWorld example?

If you're talking about this PR as a whole, then: I tested it with both root and non-root images, and it works. The change is supposed to be minimal, and it doesn't seem to affect existing workloads.

But anyway, this PR should be considered deprecated (I'll add a DONTMERGE note now), and I would like for someone to create a new PR with @woju's proposal, see other comment.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/130)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @aneessahib, @kailun-qin, and @woju)

___
*[`test/ubuntu18.04-hello-world.dockerfile` line 5 at r1](https://reviewable.io/reviews/gramineproject/gsc/130#-NPcHIAd9Fy-g0MyDPL5:-NPtZhZc0bwseumx5M6u:b-8kartn) ([raw file](https://github.com/gramineproject/gsc/blob/dfd72177c560119c26ef41325452af04d25d8870/test/ubuntu18.04-hello-world.dockerfile#L5)):*

Sorry @mkow I don't understand the question. Why did you pose the question here, in the HelloWorld example?

If you're talking about this PR as a whole, then: I tested it with both root and non-root images, and it works. The change is supposed to be minimal, and it doesn't seem to affect existing workloads.

But anyway, this PR should be considered deprecated (I'll add a DONTMERGE note now), and I would like for someone to create a new PR with @woju's proposal, see other comment.
</blockquote></details>

Ah, I misunderstood what the second commit does, resolving.

dimakuv:
Incorrect attempt at fixing the issue, closing this PR.

PR #149, opened 2023-05-18
option for not traversing links

max-lepikhin:
## Adds option for not traversing symbolic links in finalize_manifest.py.

The reason: some docker images contain many symbolic links as result of build with [bazel](https://bazel.io/). Our start-up time for an enclave with a python binary increased from ~1 minute to ~30 minutes because the manifest now contains all resolved symblinks totaling ~500K files. Caused by this change:
https://github.com/gramineproject/gsc/commit/c374329590c77ad1725e5767900a0e4679dd2215

### DIFF
--- a/finalize_manifest.py
+++ b/finalize_manifest.py
@@ -31,7 +31,7 @@ def extract_files_from_user_manifest(manifest):
     return files
 
 
-def generate_trusted_files(root_dir, already_added_files):
+def generate_trusted_files(root_dir, already_added_files, followlinks):
     excluded_paths_regex = (r'^/('
                                 r'boot/.*'
                                 r'|\.dockerenv'
@@ -55,7 +55,7 @@ def generate_trusted_files(root_dir, already_added_files):
 
     # WARNING: below loop does not remove cycles (which can happen due to symlinks), but we assume
     # that Docker images (excluding paths/files listed in `excluded_paths_regex`) don't have cycles.
-    for dirpath, dirnames, files in os.walk(root_dir.encode('UTF-8'), followlinks=True):
+    for dirpath, dirnames, files in os.walk(root_dir.encode('UTF-8'), followlinks=followlinks):
         scandirs = []
         for dirname in dirnames:
             to_be_walked_dir = os.path.join(dirpath, dirname)
@@ -127,6 +127,10 @@ def generate_library_paths():
 argparser = argparse.ArgumentParser()
 argparser.add_argument('-d', '--dir', default='/',
     help='Search directory tree from this root to generate list of trusted files.')
+argparser.add_argument('-l', '--links', action='store_true',
+    help='Whether to traverse symbolic links, default is True.')
+argparser.add_argument('-nl', '--no-links', dest='links', action='store_false')
+argparser.set_defaults(links=True)
 
 def main(args=None):
     args = argparser.parse_args(args[1:])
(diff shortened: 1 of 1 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/149)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @max-lepikhin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/149#-NVnqUgz8KB9RHUZcKoH:-NVnqUgz8KB9RHUZcKoI:b-x0x0iu) (no related file):*
How does this PR help you? Are you using your own Dockerfiles?

Because in this PR, you didn't modify this caller of `finalize_manifest.py`: https://github.com/gramineproject/gsc/blob/a524c64a57ef0cf14136b9965384645e1071b066/templates/Dockerfile.common.build.template#L61

So as is, this PR doesn't change anything in current GSC behavior...

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/149)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @max-lepikhin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/149#-NVnqmxV0Kv7IDbtr2R-:-NVnqmxV0Kv7IDbtr2R0:bmwlsh5) (no related file):*
I don't like this solution.

We generally have a problem with symlinks: https://github.com/gramineproject/gramine/issues/1176. We still don't know how to operate on them better.

___

max-lepikhin:
Yes, we'll use a custom docker file. Alternatively, we can override/patch
the python file during build. What solution would you recommend instead?

dimakuv (commented):
> Yes, we'll use a custom docker file. Alternatively, we can override/patch the python file during build. What solution would you recommend instead?

Yep, I would recommend you to patch the Python file during build. In other words, it will be a change on your side, not in the GSC repo.

On the other hand, why do you even track the GSC repo and extract some files? Wouldn't it be easier for you to just copy-paste the `finalize_manifest.py` file into your own repo and modify as you see fit?

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/149#-:-NVnx1Ba1KhtU_-HopD3:b-3w3fiu)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @max-lepikhin)

max-lepikhin:
We did it that way first. There is benefit in open source presumably, the
code will evolve in gsc repo, the two files will diverge. If you don't see
other alternatives, please feel free to reject the PR.

dimakuv (commented):
Maybe this idea will be of help to you: https://github.com/gramineproject/gsc/pull/130#pullrequestreview-1306985696

If it is enough for you to exclude some directories (where Bazel puts the most symlinks), then you won't need the hack you introduced in this PR.

If you agree that this idea should be implemented in GSC and will be of great help to you, then we can start implementing it.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/149#-:-NVo-h52BIQ8LA0vb2Hs:b-ift3lb)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @max-lepikhin)

Issue #153, opened 2023-06-16
GSC Build gets failed for latest Debian 12 & Ubuntu 23.04


anjalirai-intel:
## Description of the problem
**GSC Build gets failed for latest Debian 12 & Ubuntu 23.04**

Debian 12 & Ubuntu 23.04 has adopted PEP668 (PEP = Python Enhancement Proposal), marking the Python base environments as "externally managed", and no longer allowing regular pip install usage for both user and system installations. The default Python version which gets installed is Python 3.11. So if we try to install package using pip it throws below error:

```
/usr/bin/python3 -B -m pip install 'tomli>=1.1.0' 'tomli-w>=0.4.0' 'meson>=0.56'
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.11/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

## Steps to reproduce

1. Pull latest Debian docker image
2. Create a docker container with entrypoint bash
3. Follow the steps 4-26 from templates/debian/Dockerfile.compile.template
https://github.com/gramineproject/gsc/blob/master/templates/debian/Dockerfile.compile.template#L4..L26

## Actual results
```
apt-get install -y autoconf  bison  build-essential  coreutils  curl  gawk  git  libcurl4-openssl-dev  libprotobuf-c-dev  linux-headers-generic  nasm  ninja-build  pkg-config  protobuf-c-compiler  protobuf-compiler  python3  python3-cryptography  python3-pip  python3-protobuf  wget
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'linux-headers-amd64' instead of 'linux-headers-generic'
coreutils is already the newest version (9.1-1).
coreutils set to manually installed.
The following additional packages will be installed:
[...]

PR #155, opened 2023-07-10
Add Debian12, Ubuntu22.04 and remove Ubuntu18.04 support

amathew3:
This PR fixes the build issues seen with newer distros ubuntu22.04, debian12 etc.  For newer distros the `apt-get install` is used to install the python packages. 
Also removes the support for ubuntu18.04.

This PR fixes #153

Take the hello-world example in the test folder and modify it to use the ubuntu22.04 base image.
Edit the `config.yaml` file to select the distro as ubuntu:22.04
Do the `gsc build` and `gsc sign` commands.
Expected result: The build build should pass with out errors.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gsc/155)

PR #158, opened 2023-07-11
[Docs] Pin to specific version of Python in the example

dimakuv:
## Description of the changes 

Python default (`latest`) Docker image recently switched to Debian 12 which uses PEP668 (marking Python base environments as "externally managed"). This change is not yet supported by GSC, so let's pin the Python example to a supported version.

See also https://github.com/gramineproject/gsc/issues/153 and https://peps.python.org/pep-0668/

## How to test this PR? 

Manually test that Python example works.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gsc/158)

### DIFF
--- a/Documentation/index.rst
+++ b/Documentation/index.rst
@@ -469,29 +469,30 @@ This example assumes that all prerequisites are installed and configured.
 
       openssl genrsa -3 -out enclave-key.pem 3072
 
-#. Pull public Python image from Dockerhub:
+#. Pull public Python image from Dockerhub (we pin to the Debian Bullseye
+   version):
 
    .. code-block:: sh
 
-      docker pull python
+      docker pull python:bullseye
 
 #. Graminize the Python image using :command:`gsc build`:
 
    .. code-block:: sh
 
-      ./gsc build --insecure-args python test/generic.manifest
+      ./gsc build --insecure-args python:bullseye test/generic.manifest
 
 #. Sign the graminized Docker image using :command:`gsc sign-image`:
 
    .. code-block:: sh
 
-      ./gsc sign-image python enclave-key.pem
+      ./gsc sign-image python:bullseye enclave-key.pem
 
 #. Retrieve SGX-related information from graminized image using :command:`gsc info-image`:
 
    .. code-block:: sh
 
-      ./gsc info-image gsc-python
+      ./gsc info-image gsc-python:bullseye
 
 #. Test the graminized Docker image (change ``--device=/dev/sgx_enclave`` to
    your version of the Intel SGX driver if needed):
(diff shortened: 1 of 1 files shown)


kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/158)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/158)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/gramineproject/gsc/158#-N_3hJKWEvXnDY4C9J5p):*
How hard would it be to fix the issue instead of pinning to an older version? Using older software is a bit fishy practice :) And also easy to forget to update the version later, so if the real fix isn't too hard then I'd prefer it instead.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/158)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @woju)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/gramineproject/gsc/158#-N_3hJKWEvXnDY4C9J5p):*

How hard would it be to fix the issue instead of pinning to an older version? Using older software is a bit fishy practice :) And also easy to forget to update the version later, so if the real fix isn't too hard then I'd prefer it instead.
</blockquote></details>

The proper fix is here: https://github.com/gramineproject/gsc/pull/155

In other words, to support the latest Python version we need Debian 12 support, which in turn needs a resolution of the PEP668 problem.

However, I don't think #155 is trivial, and we'll need to wait for @woju to give his opinion... So I created this PR to go with the workaround, until the PEP668 issue is fixed.

Also, isn't it generally a good practice to pin versions of software?

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/158)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/gramineproject/gsc/158#-N_3hJKWEvXnDY4C9J5p):*
Ok then.

> Also, isn't it generally a good practice to pin versions of software?

Hmm, I think it depends on the specific case. It gives you more confidence that something will keep working and won't break spontaneously, but you may also miss important updates if that specific component isn't actively maintained.

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/158)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/158)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/158)* status: :shipit: complete! all files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/158#-N_44ysmEndKs9S5QcsQ:-N_44ysmEndKs9S5QcsR:b-4xc48i) (no related file):*
Did a final rebase and manually verified that Python example works.

___

PR #173, opened 2023-09-29
Add support for password-protected signing key

### LINKED ISSUE #148: Fix `passphrase` signing in GSC
After https://github.com/gramineproject/gramine/pull/1197 is merged, there should be a corresponding PR in GSC repo, that would:

-   Remove [https://github.com/gramineproject/gsc/blob/master/sign.sh](https://github.com/gramineproject/gsc/blob/master/sign.sh)
-   Remove `passphrase` arg from GSC itself: [https://github.com/gramineproject/gsc/blob/a60a4991d7351c1b7489444c694638ca9cc68e39/gsc.py#L504](https://github.com/gramineproject/gsc/blob/a60a4991d7351c1b7489444c694638ca9cc68e39/gsc.py#L504)
-   Fix around here: [https://github.com/gramineproject/gsc/blob/a60a4991d7351c1b7489444c694638ca9cc68e39/gsc.py#L363](https://github.com/gramineproject/gsc/blob/a60a4991d7351c1b7489444c694638ca9cc68e39/gsc.py#L363)
-   Fix around here: [https://github.com/gramineproject/gsc/blob/b7f0576219321c1b7d94b29661274419b009fc49/templates/Dockerfile.common.sign.template#L8](https://github.com/gramineproject/gsc/blob/b7f0576219321c1b7d94b29661274419b009fc49/templates/Dockerfile.common.sign.template#L8)

jkr0103:
## Description of the changes 

Now that core Gramine's `gramine-sgx-sign` tool supports the `--passphrase` argument, we don't need the `expect` tool.

Fixes #148 

Prerequisite: https://github.com/gramineproject/gramine/pull/1581

## How to test this PR? 
Use protected signing key for signing the GSC image for any workload. make sure to supply `passphrase` with signing command.

Examples:
Create protected signing key: `openssl genrsa -3 -aes128 -passout pass:test@123 -out enclave-key.pem 3072`

Sign GSC image: `./gsc sign-image <image_name> enclave-key.pem --passphrase test@123`

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)


dimakuv (changes requested):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jkr0103)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjLGWE6bWPYVW8SEjq:-NfjLGWE6bWPYVW8SEjr:b-hgmjgl) (no related file):*
I'm very happy to remove that `expect` script and dependency! Thanks for this PR.

___
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKWax0kY8AUGnKtxz):*
`password-protected`
___
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjLPcICasFCiaOVbpB):*
I would add a commit body message like this:
```
Now that core Gramine's `gramine-sgx-sign` tool supports
the `--passphrase` argument, we don't need to `expect` tool.
```
___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-NfjKtMi0TTR9AJ0Tc-r:b-w0ecvd) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*
> ```
>       --manifest /gramine/app_files/entrypoint.manifest \
>       --output /gramine/app_files/entrypoint.manifest.sgx \
>       --passphrase $passphrase
> ```

What if there is no passphrase (i.e. a classic `./gsc sign-image <image_name> enclave-key.pem`)? Won't this `gramine-sgx-sign` invocation break? Can you test this scenario?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @jkr0103)

___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-NfkO9Me9w_HJ4n4c-bl:bt1jb9q) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*

What if there is no passphrase (i.e. a classic `./gsc sign-image <image_name> enclave-key.pem`)? Won't this `gramine-sgx-sign` invocation break? Can you test this scenario?
</blockquote></details>

Oops. Your question made me realise that gramineproject/gramine#1581 is wrong. Because if `passphrase` is empty `str`, then it is returned directly (`passphrase and passphrase.encode()` is `'' and ''.encode()`, so `''`, because left side of `and` operator is false-ish; see https://docs.python.org/3/reference/expressions.html#boolean-operations). It will traceback for the same reason as in PR description.

I'm sorry for my lack of imagination while speed-drafting #1581.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKWax0kY8AUGnKtxz):*

`password-protected`
</blockquote></details>

Done.
___
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjLPcICasFCiaOVbpB):*

I would add a commit body message like this:
```
Now that core Gramine's `gramine-sgx-sign` tool supports
the `--passphrase` argument, we don't need to `expect` tool.
```
</blockquote></details>

Done.
___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-NfoI7mD0TcrgThdLOMw:b-xxhi3a) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*

Oops. Your question made me realise that gramineproject/gramine#1581 is wrong. Because if `passphrase` is empty `str`, then it is returned directly (`passphrase and passphrase.encode()` is `'' and ''.encode()`, so `''`, because left side of `and` operator is false-ish; see https://docs.python.org/3/reference/expressions.html#boolean-operations). It will traceback for the same reason as in PR description.

I'm sorry for my lack of imagination while speed-drafting #1581.
</blockquote></details>

Indeed `--passphrase` arg needs non-empty password otherwise it complains `Error: Option '--passphrase' requires an argument.`

Are we going to fix it in Gramine? otherwise I have to add `if else` contition and add `--passphrase` based on password length.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jkr0103 and @woju)

___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-NfoMAofEWi_xjVkRMii:bvk8s6b) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*

Indeed `--passphrase` arg needs non-empty password otherwise it complains `Error: Option '--passphrase' requires an argument.`

Are we going to fix it in Gramine? otherwise I have to add `if else` contition and add `--passphrase` based on password length.
</blockquote></details>

@woju I think it shouldn't be a big problem? I mean, an empty-string argument in command-line args is something very unexpected, and I don't think it's too common to merit a separate handling for this case.

@jkr0103 I think it's better to fix it here, in GSC. As you said, an `if else` condition should be good.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: 6 of 7 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-Nfp0Tkp8uG3mzSS7n-4:blovlft) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*

@woju I think it shouldn't be a big problem? I mean, an empty-string argument in command-line args is something very unexpected, and I don't think it's too common to merit a separate handling for this case.

@jkr0103 I think it's better to fix it here, in GSC. As you said, an `if else` condition should be good.
</blockquote></details>

I have fixed it in GSC

dimakuv (changes requested):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @woju)

___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-NfqidwX02L3LMh0Urpc:b-r4v6fl) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*

I have fixed it in GSC
</blockquote></details>

Hm, looks complex. Can't you do Bash parameter substitution:
```
RUN {% block path %}{% endblock %} gramine-sgx-sign \
      --key /gramine/app_files/gsc-signer-key.pem \
      --manifest /gramine/app_files/entrypoint.manifest \
      --output /gramine/app_files/entrypoint.manifest.sgx \
      ${passphrase:+--passphrase $passphrase}
```

You won't need `passphrase_name` at all then.

@woju Is the above a right POSIX shell kung-fu?

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-NftEK-41uNo0kX0Yf73:b-qpwjgw) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*

Hm, looks complex. Can't you do Bash parameter substitution:
```
RUN {% block path %}{% endblock %} gramine-sgx-sign \
      --key /gramine/app_files/gsc-signer-key.pem \
      --manifest /gramine/app_files/entrypoint.manifest \
      --output /gramine/app_files/entrypoint.manifest.sgx \
      ${passphrase:+--passphrase $passphrase}
```

You won't need `passphrase_name` at all then.

@woju Is the above a right POSIX shell kung-fu?
</blockquote></details>

`${passphrase:+--passphrase $passphrase}` works fine. Shell I make the change in PR?

And `sgx-sign` tool also support `--password` along with `-p` and `--passphrase` here: https://github.com/gramineproject/gramine/blob/94899398023e2dd2b9486aac03cc2996c2f1b13d/python/graminelibos/sgx_sign.py#L591 but GSC doesn't, code: https://github.com/gramineproject/gsc/blob/5099b641e171942f87254d751fe4bee086a869c2/gsc.py#L547


Can I make same change in GSC?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @woju)

___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-NftN84oEpp3ulxLkUG9:b-nh9trl) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*
> Shell I make the change in PR?

Yep, I'm 95% sure that it's a correct and portable way. Still would like to know @woju's opinion, but to me this approach looks correct.

> Can I make same change in GSC?

Yes, I think it makes sense to have Gramine and GSC use/allow the same options.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-NfzfAdA05pWhTdmur6v:bpa9qhi) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*

> Shell I make the change in PR?

Yep, I'm 95% sure that it's a correct and portable way. Still would like to know @woju's opinion, but to me this approach looks correct.

> Can I make same change in GSC?

Yes, I think it makes sense to have Gramine and GSC use/allow the same options.
</blockquote></details>

@woju Could you provide your inputs?

woju (commented):
> Reviewed 6 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/173#-NgxSjqPEb0hnU_s5je6:-NgxSjqPEb0hnU_s5je7:b-r39a4l) (no related file):*
I've created gramineproject/gramine#1611 which fixes the case for empty passphrase. It can be merged before or after this PR, in case someone elects to use empty (but set) passwords.

___
___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-NgxPpcbBjTi9zKQ60_D:brnlxzb) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*
`${passphrase+--passphrase "$passphrase"}` if you'd like shell-fu — no colon before plus and add quotes inside `{}` and don't add quotes outside of `{}`, I believe, in case passphrase is either set but empty or contains whitespace.

> empty-string argument in command-line args is something very unexpected

It's common enough that we can't discount it, but I argue this needs to be fixed in Gramine proper (#1611). GSC should just pass through.

Test for gsc's `--passphrase`:
- set and non-empty (should result in `gramine-sgx-sign ... --passphrase "abc"`)
- set but empty (should result in `gramine-sgx-sign ... --passphrase ""`)
- unset (should result in `gramine-sgx-sign ...` — no `--passphrase added at all)

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02, and we're interested in "set but null" case in the table therein.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-Nh0W0vq9_FFkstOKraO:bv4gu50) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*

`${passphrase+--passphrase "$passphrase"}` if you'd like shell-fu — no colon before plus and add quotes inside `{}` and don't add quotes outside of `{}`, I believe, in case passphrase is either set but empty or contains whitespace.

> empty-string argument in command-line args is something very unexpected

It's common enough that we can't discount it, but I argue this needs to be fixed in Gramine proper (#1611). GSC should just pass through.

Test for gsc's `--passphrase`:
- set and non-empty (should result in `gramine-sgx-sign ... --passphrase "abc"`)
- set but empty (should result in `gramine-sgx-sign ... --passphrase ""`)
- unset (should result in `gramine-sgx-sign ...` — no `--passphrase added at all)

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02, and we're interested in "set but null" case in the table therein.
</blockquote></details>

I tried `${passphrase+--passphrase "$passphrase"}` and it doesn't work without colon. Error below:



_Code snippet:_
```
  File "/gramine/meson_build_output/lib/python3.9/site-packages/graminelibos/sgx_sign.py", line 595, in sign_with_file

    private_key = load_private_key_from_pem_file(key, passphrase and passphrase.encode())
  File "/gramine/meson_build_output/lib/python3.9/site-packages/graminelibos/sgx_sign.py", line 608, in load_private_key_from_pem_file

    private_key = serialization.load_pem_private_key(
  File "/usr/lib/python3/dist-packages/cryptography/hazmat/primitives/serialization/base.py", line 18, in load_pem_private_key
    return backend.load_pem_private_key(data, password)
  File "/usr/lib/python3/dist-packages/cryptography/hazmat/backends/openssl/backend.py", line 1244, in load_pem_private_key

    return self._load_key(
  File "/usr/lib/python3/dist-packages/cryptography/hazmat/backends/openssl/backend.py", line 1447, in _load_key

    utils._check_byteslike("password", password)
  File "/usr/lib/python3/dist-packages/cryptography/utils.py", line 36, in _check_byteslike
    raise TypeError("{} must be bytes-like".format(name))
TypeError: password must be bytes-like
```

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @woju)

___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-Nh0Zzdi0KI5qqIIMRC8:b-u3dzvt) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*

I tried `${passphrase+--passphrase "$passphrase"}` and it doesn't work without colon. Error below:
</blockquote></details>

@woju I think we can't use he `+` version (that substitutes word if parameter is Set But Null). That's because IIUC in Docker, the `ARG passphrase` line literally creates a `passphrase` variable, i.e. it is never NULL (but empty if there was no passphrase given to the Docker build).

So I think we should stick to the `:+` version.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: 5 of 7 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-NhQWWZZ2kdcjD9NTOdj:b7gdy3t) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*

@woju I think we can't use he `+` version (that substitutes word if parameter is Set But Null). That's because IIUC in Docker, the `ARG passphrase` line literally creates a `passphrase` variable, i.e. it is never NULL (but empty if there was no passphrase given to the Docker build).

So I think we should stick to the `:+` version.
</blockquote></details>

I have updated the PR as validation need to start ASAP. @woju please provide your inputs if latest changes still need some change.

@[anjalirai-intel](https://github.com/anjalirai-intel) you may start validating the PR.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

woju (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[`templates/Dockerfile.common.sign.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/173#-NfjKtMi0TTR9AJ0Tc-q:-NhR4aaOEEVIL0e_IOuY:b-3ru392) ([raw file](https://github.com/gramineproject/gsc/blob/aa24df6e3cc3c547cbdbdef621728985e7468241/templates/Dockerfile.common.sign.template#L10)):*
> [@woju](https://github.com/woju) I think we can't use he `+` version (that substitutes word if parameter is Set But Null). That's because IIUC in Docker, the `ARG passphrase` line literally creates a `passphrase` variable, i.e. it is never NULL (but empty if there was no passphrase given to the Docker build).

If so, then yes, `:+` is the correct thing to do. I'm resolving.

> Error below:

This should have been fixed by gramineproject/gramine#1611.

mkow (changes requested):
> Reviewed 5 of 7 files at r1, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @jkr0103)

___
*[`templates/Dockerfile.common.sign.template` line 11 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-NhTodO3E96DEBLLlIik:-NhTodO3E96DEBLLlIil:b-yustm2) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L11)):*
> ```
>       --manifest /gramine/app_files/entrypoint.manifest \
>       --output /gramine/app_files/entrypoint.manifest.sgx \
>       ${passphrase:+--passphrase "$passphrase"}
> ```

Does this work if the password contains `"` or `\`?
Also, saving user key's password to a file on the disk doesn't seem like a good practice.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @jkr0103)

___
*[`templates/Dockerfile.common.sign.template` line 11 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-NhTodO3E96DEBLLlIik:-NhTpFfX3CkAy8ZCSfCg:b656yvn) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L11)):*
> Also, saving user key's password to a file on the disk doesn't seem like a good practice.

Hmm, or maybe not? How does this line actually work?

// Regardless of this, the question about special chars still stands.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @jkr0103)

___
*[`templates/Dockerfile.common.sign.template` line 11 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-NhTodO3E96DEBLLlIik:-NhVXfQU3oLhsdiaHxXN:bqpreqc) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L11)):*
> > Also, saving user key's password to a file on the disk doesn't seem like a good practice.
> 
> Hmm, or maybe not? How does this line actually work?

Why do you think there's anything about files in this line? It's a bash/shell substitution syntax, see the table here:  https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02. The password string is passed as an input argument everywhere here.

> Does this work if the password contains `"` or `\`?

Good question. I guess @jkr0103 will have to experiment with such passphrases. And report the results. @mkow If these symbols need to be escaped during `gsc build --passphrase <...>`, would it be enough to add this note in the GSC documentation?

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @jkr0103)

___
*[`templates/Dockerfile.common.sign.template` line 11 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-NhTodO3E96DEBLLlIik:-NhZ7woZCzrMeUNlAMQb:b-et15cv) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L11)):*
> Why do you think there's anything about files in this line? It's a bash/shell substitution syntax, see the table here: [https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3\_chap02.html#tag\_18\_06\_02](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02). The password string is passed as an input argument everywhere here.

Ok, I've read more about this and it's not Bash and not Jinja, but Docker doing the substitution here before running the command.

btw. quoting Dockerfile `ARG` documentation:

> Warning:
> It is not recommended to use build-time variables for passing secrets like GitHub keys, user credentials etc. Build-time variable values are visible to any user of the image with the `docker history` command.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @jkr0103)

___
*[`templates/Dockerfile.common.sign.template` line 11 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-NhTodO3E96DEBLLlIik:-Nh_b5RiBri5JDGfowKd:b-q4838a) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L11)):*
> Ok, I've read more about this and it's not Bash and not Jinja, but Docker doing the substitution here before running the command.

Where did you read this?

> btw. quoting Dockerfile `ARG` documentation:

But that's why we have a trick of `FROM ...` (a multi-stage build), which removes all files and all ENVs and ARGs from the previous image, see my other comment where I highlight this place.
___
*[`templates/Dockerfile.common.sign.template` line 15 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-Nh_bInN2WXx37AB3mr4:-Nh_bInN2WXx37AB3mr5:b-o4o4wa) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L15)):*
> ```
> # This trick removes all temporary files from the previous commands (including gsc-signer-key.pem
> # and passphrase)
> FROM {{image}}
> ```

To @mkow: this trick removes the previous `ARG` from the final signed image.

To make sure myself, I ran this PR on my machine (as described in the PR description, with a passphase-protected key) and did:
```
$ docker history --no-trunc gsc-python:bullseye | grep test@123
< returns nothing >
```

Looking at `docker history`, it's also clear that the steps in-between two `FROM` keywords are "vanished":
```
$ docker history gsc-python:bullseye
IMAGE          CREATED          CREATED BY                                      SIZE      COMMENT
21dfb8f165fe   10 minutes ago   /bin/sh -c #(nop) COPY file:6c832ef8e306f999…   9.35MB
d3d03dbd709c   10 minutes ago   /bin/sh -c #(nop) COPY file:ba82810b1d6c2ef4…   1.81kB
340188eab1f0   19 minutes ago   /bin/sh -c #(nop)  ENTRYPOINT ["/bin/bash" "…   0B
e75637d26da0   19 minutes ago   |0 /bin/sh -c chmod u+x /gramine/app_files/a…   4.19MB
...
```

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`templates/Dockerfile.common.sign.template` line 11 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-NhTodO3E96DEBLLlIik:-NhazyYY0f5qQ5S8-oEJ:bha8r9q) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L11)):*
>Does this work if the password contains `"` or `\`?

I tried with `"` and `\` as below and that worked:

openssl genrsa -3 -aes128 -passout pass:"  test@123" -out enclave-key.pem 3072
./gsc sign-image --password "  test@123" redis:7.0.10 enclave-key.pem

openssl genrsa -3 -aes128 -passout pass:\"test@123 -out enclave-key.pem 3072
./gsc sign-image --password \"test@123 redis:7.0.10 enclave-key.pem

openssl genrsa -3 -aes128 -passout pass:\\test@123 -out enclave-key.pem 3072
./gsc sign-image --password \\test@123 redis:7.0.10 enclave-key.pem

woju (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`templates/Dockerfile.common.sign.template` line 11 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-NhTodO3E96DEBLLlIik:-NhbAwIy-9MfGxXP_UtJ:b-iksxcb) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L11)):*

>Does this work if the password contains `"` or `\`?

I tried with `"` and `\` as below and that worked:

openssl genrsa -3 -aes128 -passout pass:"  test@123" -out enclave-key.pem 3072
./gsc sign-image --password "  test@123" redis:7.0.10 enclave-key.pem

openssl genrsa -3 -aes128 -passout pass:\"test@123 -out enclave-key.pem 3072
./gsc sign-image --password \"test@123 redis:7.0.10 enclave-key.pem

openssl genrsa -3 -aes128 -passout pass:\\test@123 -out enclave-key.pem 3072
./gsc sign-image --password \\test@123 redis:7.0.10 enclave-key.pem
</blockquote></details>

`\` and `"` will work in `"$passphrase"` if docker does correct thing in the setting of this variable in the environment. Same as in plain (non-docker) shell, if `"` is already in the content of the variable, then substituting the variable wrapped in `"` changes nothing, because the content of variable is considered already "quoted" and the `"` quotes surrounding the substitution are not, so they're removed (after word splitting). Same for `\`, it's already quoted.

POSIX shell is almost dumb string replacement, but not quite. To have complete mental model, you need to know which characters are considered quoted and how word splitting works. Here it's correct.

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`templates/Dockerfile.common.sign.template` line 15 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-Nh_bInN2WXx37AB3mr4:-Nhb5ljI7rLtTUyGcCy_:be8kcvz) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L15)):*

To @mkow: this trick removes the previous `ARG` from the final signed image.

To make sure myself, I ran this PR on my machine (as described in the PR description, with a passphase-protected key) and did:
```
$ docker history --no-trunc gsc-python:bullseye | grep test@123
< returns nothing >
```

Looking at `docker history`, it's also clear that the steps in-between two `FROM` keywords are "vanished":
```
$ docker history gsc-python:bullseye
IMAGE          CREATED          CREATED BY                                      SIZE      COMMENT
21dfb8f165fe   10 minutes ago   /bin/sh -c #(nop) COPY file:6c832ef8e306f999…   9.35MB
d3d03dbd709c   10 minutes ago   /bin/sh -c #(nop) COPY file:ba82810b1d6c2ef4…   1.81kB
340188eab1f0   19 minutes ago   /bin/sh -c #(nop)  ENTRYPOINT ["/bin/bash" "…   0B
e75637d26da0   19 minutes ago   |0 /bin/sh -c chmod u+x /gramine/app_files/a…   4.19MB
...
```
</blockquote></details>

You are right that this will clear the password from the final image, because of "FROM" command.

IMHO it's still not a good solution because you are saving a password to a PEM file in the `unsigned_image` which means that you are leaking passwords all over the signing machine. Also while signing another user might see your password in `ps(1)` command (so in general I'm not a fun of `--passphrase` option). If you are leaking this password, why is PEM protected with a password at all?

Maybe you can consider using `RUN --mount=type=secret` from buildx? But this doesn't allow to do stdin IMHO. https://render.com/docs/docker-secrets
Or use docker secrets from swarm? https://docs.docker.com/engine/swarm/secrets/ I guess you could add a temporary secret in gsc, build it, and then remove it. Or require a user to create secret and then just accept the name of the secret.

The more hardcore solution, which I have suggested in a different project, would be to do a fake "multistage" build.
When you finish building `unsigned_image`, then run it to do `gramine-sgx-sign` and use the output of these as input to build another Dockerfile.
Then the passphrase could be passed as stdin.

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`templates/Dockerfile.common.sign.template` line 15 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-Nh_bInN2WXx37AB3mr4:-NhbEP7G2bPDXhes_1gC:bui44ol) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L15)):*
<details><summary><i>Previously, oshogbo (Mariusz Zaborski) wrote…</i></summary><blockquote>

You are right that this will clear the password from the final image, because of "FROM" command.

IMHO it's still not a good solution because you are saving a password to a PEM file in the `unsigned_image` which means that you are leaking passwords all over the signing machine. Also while signing another user might see your password in `ps(1)` command (so in general I'm not a fun of `--passphrase` option). If you are leaking this password, why is PEM protected with a password at all?

Maybe you can consider using `RUN --mount=type=secret` from buildx? But this doesn't allow to do stdin IMHO. https://render.com/docs/docker-secrets
Or use docker secrets from swarm? https://docs.docker.com/engine/swarm/secrets/ I guess you could add a temporary secret in gsc, build it, and then remove it. Or require a user to create secret and then just accept the name of the secret.

The more hardcore solution, which I have suggested in a different project, would be to do a fake "multistage" build.
When you finish building `unsigned_image`, then run it to do `gramine-sgx-sign` and use the output of these as input to build another Dockerfile.
Then the passphrase could be passed as stdin.
</blockquote></details>

Actually, my hardcore idea could be simplified to:
```
docker build -t Dockerfile .
docker run --mountpoints=... --entrypoint=gramine-sign --key ... --manifest .. --output ...
docker commit
```

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`templates/Dockerfile.common.sign.template` line 11 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-NhTodO3E96DEBLLlIik:-NhbGLhq0QHfrz8Pj0CJ:b-7tdf7d) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L11)):*

`\` and `"` will work in `"$passphrase"` if docker does correct thing in the setting of this variable in the environment. Same as in plain (non-docker) shell, if `"` is already in the content of the variable, then substituting the variable wrapped in `"` changes nothing, because the content of variable is considered already "quoted" and the `"` quotes surrounding the substitution are not, so they're removed (after word splitting). Same for `\`, it's already quoted.

POSIX shell is almost dumb string replacement, but not quite. To have complete mental model, you need to know which characters are considered quoted and how word splitting works. Here it's correct.
</blockquote></details>

agree
___
*[`templates/Dockerfile.common.sign.template` line 15 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-Nh_bInN2WXx37AB3mr4:-NhbGPd88wGFpGB2ahJ2:bq6bt2p) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L15)):*
<details><summary><i>Previously, oshogbo (Mariusz Zaborski) wrote…</i></summary><blockquote>

Actually, my hardcore idea could be simplified to:
```
docker build -t Dockerfile .
docker run --mountpoints=... --entrypoint=gramine-sign --key ... --manifest .. --output ...
docker commit
```
</blockquote></details>

The final signed GSC image doesn't leak the password but the dangling image generated as part of multistage build does leak the password and password can be retrieve using `docker history <dangling_image_id>`. If we can delete this dangling image than we should be good. I am about to create PR for deleting this dangling image.

jkr0103 (commented):
> Reviewed 1 of 7 files at r1, 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`templates/Dockerfile.common.sign.template` line 15 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-Nh_bInN2WXx37AB3mr4:-NhbLTs_7PHZjkB1R4Lc:b-jhyt44) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L15)):*

The final signed GSC image doesn't leak the password but the dangling image generated as part of multistage build does leak the password and password can be retrieve using `docker history <dangling_image_id>`. If we can delete this dangling image than we should be good. I am about to create PR for deleting this dangling image.
</blockquote></details>

I have create PR https://github.com/gramineproject/gsc/pull/175 for deleting the dangling image generated during GSC signing pocess.

aneessahib (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`templates/Dockerfile.common.sign.template` line 15 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-Nh_bInN2WXx37AB3mr4:-Nhe54vT9Rr0Cab01_kX:bqd3mxv) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L15)):*

I have create PR https://github.com/gramineproject/gsc/pull/175 for deleting the dangling image generated during GSC signing pocess.
</blockquote></details>

@oshogbo @woju Agreed- FROM trick wouldn't really help if the signing machine is compromised with wrong users authenticated to use it. It only helps in the finally deployed image to not carry secrets (this is how we take out the signing key itself out of the final image since the founding days of GSC). Can we continue with the same security scope for now and while we work separately on hardening solutions for the build phase?

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

kailun-qin (approved):
> Reviewed 5 of 7 files at r1, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[`templates/Dockerfile.common.sign.template` line 15 at r4](https://reviewable.io/reviews/gramineproject/gsc/173#-Nh_bInN2WXx37AB3mr4:-Nhjl-oI4Ae5ZYkA6dbu:bknq6xi) ([raw file](https://github.com/gramineproject/gsc/blob/8606fce9344d98b3f95f7e8b49d512517221f686/templates/Dockerfile.common.sign.template#L15)):*

@oshogbo @woju Agreed- FROM trick wouldn't really help if the signing machine is compromised with wrong users authenticated to use it. It only helps in the finally deployed image to not carry secrets (this is how we take out the signing key itself out of the final image since the founding days of GSC). Can we continue with the same security scope for now and while we work separately on hardening solutions for the build phase?
</blockquote></details>

I think what @aneessahib said is a consensus for now. And we hope for @oshogbo to submit a PR to fix all those issues with correctly using secret files/passwords :)

This PR seems to be good to be merged, so I'll do it now.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/173)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #177, opened 2023-11-02
Automatically detect the distro of the supplied docker image

jkr0103:
## Description of the changes

GSC can automatically detect the distro of the supplied image, hence user don't have to update the distro in `config.yaml` after this PR.

## How to test this PR?

1. No need to make any changes in the `config.yaml` for setting distro, do the GSC build and test.
2. Set supported distro and do GSC build and test

### DIFF
--- a/Documentation/index.rst
+++ b/Documentation/index.rst
@@ -258,8 +258,8 @@ follows three main stages and produces an image named ``gsc-<image-name>``.
 
 #. **Building Gramine.** The first stage builds Gramine from sources based on
    the provided configuration (see :file:`config.yaml`) which includes the
-   distribution (e.g., Ubuntu 20.04), Gramine repository, and the Intel SGX
-   driver details. This stage can be skipped if :command:`gsc build` uses a
+   distribution, Gramine repository, and the Intel SGX driver details.
+   This stage can be skipped if :command:`gsc build` uses a
    pre-built Gramine Docker image.
 
 #. **Graminizing the application image.** The second stage copies the important
@@ -309,10 +309,12 @@ in :file:`config.yaml.template`.
 .. describe:: Distro
 
    Defines Linux distribution to be used to build Gramine in. This distro should
-   match the distro underlying the application's Docker image; otherwise the
-   results may be unpredictable. Currently supported distros are Ubuntu 20.04,
-   Ubuntu 21.04, Ubuntu 22.04, Ubuntu 23.04, Debian 10, Debian 11, Debian 12 and
-   CentOS 8. Default value is ``ubuntu:20.04``.
+   match the distro of the supplied Docker image; otherwise the results may be
+   unpredictable. Currently supported distros are Ubuntu 20.04, Ubuntu 21.04,
+   Ubuntu 22.04, Ubuntu 23.04, Debian 10, Debian 11, Debian 12 and CentOS 8.
+   Default value is ``auto`` which means GSC automatically detects the distro
+   of the supplied Docker image. User also have the option to provide one of the
+   supported distros mentioned above.
 
 .. describe:: Registry
 
--- a/config.yaml.template
+++ b/config.yaml.template
@@ -6,7 +6,10 @@
 # - ubuntu:20.04, ubuntu:21.04, ubuntu:22.04, ubuntu:23.04
 # - debian:10, debian:11, debian:12
 # - centos:8
-Distro: "ubuntu:20.04"
+
+# GSC detects the distro automatically of the supplied image if ditro is set to `auto`.
+# User also have the option to provide one of the supported distros mentioned above.
+Distro: "auto"
 
 # If the image has a specific registry, define it here.
 # Empty by default; example value: "registry.access.redhat.com/ubi8".
--- a/gsc.py
+++ b/gsc.py
@@ -9,6 +9,7 @@ import json
 import hashlib
 import os
 import pathlib
+import re
 import shutil
 import struct
 import sys
@@ -176,6 +177,37 @@ def merge_two_dicts(dict1, dict2, path=[]):
             dict1[key] = dict2[key]
     return dict1
 
+def get_image_distro(docker_socket, image_name):
+    output = docker_socket.containers.run(image_name, entrypoint='cat /etc/os-release', remove=True)
+    output = output.decode('UTF-8')
+
+    pattern_id = re.compile('^ID=(.*)', flags=re.MULTILINE)
+    match = pattern_id.search(output)
+    if match is None:
+        raise ValueError(f'Error: Could not find distro ID')
+
+    distro_id = match.group(1)
+
+    pattern_version_id = re.compile('^VERSION_ID=\"(.*)\"', flags=re.MULTILINE)
+    match = pattern_version_id.search(output)
+    if match is None:
+        raise ValueError(f'Error: Could not find distro VERSION_ID')
+
+    distro_version_id = match.group(1)
+    return distro_id + ':' + distro_version_id
+
+def fetch_and_validate_distro_support(docker_socket, image_name, env):
+    distro = env.globals['Distro']
+    if distro == 'auto':
+        distro = get_image_distro(docker_socket, image_name)
+        env.globals['Distro'] = distro
+
+    distro, _ = distro.split(':')
+    if not os.path.exists(f'templates/{distro}'):
+        raise FileNotFoundError(f'{distro} distro is not supported by GSC.')
+
+    return distro
+
 # Command 1: Build unsigned graminized Docker image from original app Docker image.
 def gsc_build(args):
     original_image_name = args.image                           # input original-app image name
@@ -215,9 +247,12 @@ def gsc_build(args):
 
     os.makedirs(tmp_build_path, exist_ok=True)
 
-    distro = env.globals['Distro']
+    try:
+        distro = fetch_and_validate_distro_support(docker_socket, original_image_name, env)
+    except Exception as e:
+        print(e, file=sys.stderr)
+        sys.exit(1)
 
-    distro, _ = distro.split(':')
     env.globals.update({'compile_template': f'{distro}/Dockerfile.compile.template'})
     env.loader = jinja2.FileSystemLoader('templates/')
 
@@ -326,9 +361,12 @@ def gsc_build_gramine(args):
 
     os.makedirs(tmp_build_path, exist_ok=True)
 
-    distro = env.globals['Distro']
+    try:
+        distro = fetch_and_validate_distro_support(docker_socket, args.image, env)
+    except Exception as e:
+        print(e, file=sys.stderr)
+        sys.exit(1)
 
-    distro, _ = distro.split(':')
     env.loader = jinja2.FileSystemLoader('templates/')
 
     # generate Dockerfile.compile from Jinja-style templates/<distro>/Dockerfile.compile.template
@@ -381,9 +419,13 @@ def gsc_sign_image(args):
     extract_user_from_image_config(unsigned_image.attrs['Config'], env)
     env.globals['args'] = extract_define_args(args)
     env.tests['trueish'] = test_trueish
-    distro = env.globals['Distro']
 
-    distro, _ = distro.split(':')
+    try:
+        distro = fetch_and_validate_distro_support(docker_socket, args.image, env)
+    except Exception as e:
+        print(e, file=sys.stderr)
+        sys.exit(1)
+
     env.loader = jinja2.FileSystemLoader('templates/')
     sign_template = env.get_template(f'{distro}/Dockerfile.sign.template')
 


dimakuv (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jkr0103)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJFlzS3GIyhfqRG3Pv:-NiJFlzS3GIyhfqRG3Pw:bvzpjqr) (no related file):*
This change looks great to me, and makes using GSC much easier!

After this PR, `config.yaml` will be basically the same for every Docker image to be graminized.

___
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJDfZj3_XnEtFtEUS0):*
`the distro` -> `OS distro`
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJDiOMARhFcenGzCql):*
`docker` -> `Docker`
___
*[`config.yaml.template` line 3 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJDsNy8DFkCCz856JQ:-NiJDsNy8DFkCCz856JR:b7e11bn) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/config.yaml.template#L3)):*
> ```
> # Specify the OS distro that is used to build Gramine, i.e., the distro from where the Gramine build
> # gets all tools and dependencies from. This distro should match the distro underlying the
> # application's Docker image; otherwise the results may be unpredictable.
> ```

Please append smth like this:
```
... otherwise the results may be unpredictable (if you specify `"auto"`, which is
recommended, you don't need to worry about the mismatch).
```
___
*[`config.yaml.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJE9phBWtOaLPt9Bap:-NiJE6iM1ZWBKU7ZlueY:b-i7kyhd) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/config.yaml.template#L10)):*
> ```
> # - centos:8
> 
> # GSC detects the distro automatically of the supplied image if ditro is set to `auto`.
> ```

Slightly rewording:
```
# If Distro is set to "auto", GSC detects the distro automatically by examining the supplied
# Docker image. Alternatively, Distro can be set to one of the supported distros mentioned
# above.
```
___
*[`gsc.py` line 197 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJFYUT0SW5j9C7Ehey:-NiJFYUT0SW5j9C7Ehez:bj68wxj) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/gsc.py#L197)):*
> ```Python
> 
>     distro_version_id = match.group(1)
>     return distro_id + ':' + distro_version_id
> ```

@woju Could I ask you to review at least this function (or better the whole PR)? Are we guaranteed to have `/etc/os-release` file in all sensible Linux-based distros, and are we guaranteed to find `ID=...` and `VERSION_ID=...` values in this file?
___
*[`Documentation/index.rst` line 316 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJErFc7VNBRIPZJ0fh:-NiJErFc7VNBRIPZJ0fi:b-500edp) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/Documentation/index.rst#L316)):*
> ```reStructuredText
>    Ubuntu 22.04, Ubuntu 23.04, Debian 10, Debian 11, Debian 12 and CentOS 8.
>    Default value is ``auto`` which means GSC automatically detects the distro
>    of the supplied Docker image. User also have the option to provide one of the
> ```

`Users`

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: 1 of 3 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJDfZj3_XnEtFtEUS0):*

`the distro` -> `OS distro`
</blockquote></details>

Done.
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJDiOMARhFcenGzCql):*

`docker` -> `Docker`
</blockquote></details>

Done.
___
*[`config.yaml.template` line 3 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJDsNy8DFkCCz856JQ:-NiJWo5w-Ql4AgrTFSZs:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/config.yaml.template#L3)):*

Please append smth like this:
```
... otherwise the results may be unpredictable (if you specify `"auto"`, which is
recommended, you don't need to worry about the mismatch).
```
</blockquote></details>

Done.
___
*[`config.yaml.template` line 10 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJE9phBWtOaLPt9Bap:-NiJWowy3NRtes4YWxli:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/config.yaml.template#L10)):*

Slightly rewording:
```
# If Distro is set to "auto", GSC detects the distro automatically by examining the supplied
# Docker image. Alternatively, Distro can be set to one of the supported distros mentioned
# above.
```
</blockquote></details>

Done.
___
*[`gsc.py` line 197 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJFYUT0SW5j9C7Ehey:-NiJWqLLBT8E9mPOGUO2:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/gsc.py#L197)):*

@woju Could I ask you to review at least this function (or better the whole PR)? Are we guaranteed to have `/etc/os-release` file in all sensible Linux-based distros, and are we guaranteed to find `ID=...` and `VERSION_ID=...` values in this file?
</blockquote></details>

Done.
___
*[`Documentation/index.rst` line 316 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJErFc7VNBRIPZJ0fh:-NiJWr0c5cIvHhupGBUL:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/Documentation/index.rst#L316)):*

`Users`
</blockquote></details>

Done.

dimakuv (commented):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @woju)

woju (changes requested):
> Reviewed 1 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @jkr0103)

___
*[`gsc.py` line 197 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJFYUT0SW5j9C7Ehey:-NiJY3Gk3GPc-IW2HBDF:b-e95904) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/gsc.py#L197)):*
> Are we guaranteed to have `/etc/os-release` file in all sensible Linux-based distros,

I'd say it's a reasonable assumption. It's a feature of systemd and "standardised" by that project: https://manpages.debian.org/os-release. It's not "guaranteed", but it's there in all distros Gramine (main project) lists as supported, because they're all either are based on systemd (all but Alpine Linux), or just have this file (AL). But you can compile Gramine on a non-systemd distro without this file just fine, so I think it's good that this file is queried only for `auto`. I understand it's a convenience feature which is OK to fault from time to time and worst case the user will just specify the distro manually. So what.

> and are we guaranteed to find `ID=...`

Same answer as above, it's there on all supported distros, but even Mr. Poettering's spec says MAY (not SHOULD nor MUST), so it's a very weak guarantee. Same recommendation as above, keep it only for explicit `"auto"`, and let users override if autodetection breaks.

> and `VERSION_ID=...` values in this file?

This field is risky, because distros are free to specify the version very precisely, like Alpine:

```
% docker run -it --rm alpine cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.17.3
PRETTY_NAME="Alpine Linux v3.17"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
```

So I guess you want to truncate it somewhere, and this depends on distro (alpine would be split at the second dot, cf. PRETTY_NAME for what is really the version).
___
*[`gsc.py` line 184 at r2](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJaSdFBOKP8NgSNa0f:-NiJaSdFBOKP8NgSNa0g:bf1qer0) ([raw file](https://github.com/gramineproject/gsc/blob/a3e19700c7669d168776b3bede4316dbbcc40719/gsc.py#L184)):*
In any case this regexp-based parser is wrong (blocking here on this), because some values might be wrapped in double quotes, but some might not be, and this changes from distro to distro for the same value. You need `shlex.split` to parse this:

```python
>>> import pprint, shlex
>>> os_release = {}
>>> with open('/etc/os-release') as file:
...     for line in file:
...         k, v = line.split('=', 1)
...         v, = shlex.split(v)
...         os_release[k] = v
... 
>>> pprint.pprint(os_release)
{'BUG_REPORT_URL': 'https://bugs.debian.org/',
 'HOME_URL': 'https://www.debian.org/',
 'ID': 'debian',
 'NAME': 'Debian GNU/Linux',
 'PRETTY_NAME': 'Debian GNU/Linux 12 (bookworm)',
 'SUPPORT_URL': 'https://www.debian.org/support',
 'VERSION': '12 (bookworm)',
 'VERSION_CODENAME': 'bookworm',
 'VERSION_ID': '12'}
```

You can do it as one-liner, if you don't care about syntactically invalid files:
```python
>>> with open('/etc/os-release') as file:
...     os_release = dict(shlex.split(line)[0].split('=') for line in file)
```

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @woju)

___
*[`gsc.py` line 197 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJFYUT0SW5j9C7Ehey:-NiKPvVf7kehaw5_5mNK:b-wjt8p8) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/gsc.py#L197)):*

> Are we guaranteed to have `/etc/os-release` file in all sensible Linux-based distros,

I'd say it's a reasonable assumption. It's a feature of systemd and "standardised" by that project: https://manpages.debian.org/os-release. It's not "guaranteed", but it's there in all distros Gramine (main project) lists as supported, because they're all either are based on systemd (all but Alpine Linux), or just have this file (AL). But you can compile Gramine on a non-systemd distro without this file just fine, so I think it's good that this file is queried only for `auto`. I understand it's a convenience feature which is OK to fault from time to time and worst case the user will just specify the distro manually. So what.

> and are we guaranteed to find `ID=...`

Same answer as above, it's there on all supported distros, but even Mr. Poettering's spec says MAY (not SHOULD nor MUST), so it's a very weak guarantee. Same recommendation as above, keep it only for explicit `"auto"`, and let users override if autodetection breaks.

> and `VERSION_ID=...` values in this file?

This field is risky, because distros are free to specify the version very precisely, like Alpine:

```
% docker run -it --rm alpine cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.17.3
PRETTY_NAME="Alpine Linux v3.17"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
```

So I guess you want to truncate it somewhere, and this depends on distro (alpine would be split at the second dot, cf. PRETTY_NAME for what is really the version).
</blockquote></details>

Thanks @woju !

@jkr0103 Can you add a comment at the `VERSION_ID` line that some OS distros (e.g. Alpine) have very precise versions (e.g. `3.17.3`), and to support these OS distros, we'll need to special case them in this function?
___
*[`gsc.py` line 184 at r2](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJaSdFBOKP8NgSNa0f:-NiKQSf68L4mj04igSGR:bpfa0cj) ([raw file](https://github.com/gramineproject/gsc/blob/a3e19700c7669d168776b3bede4316dbbcc40719/gsc.py#L184)):*

In any case this regexp-based parser is wrong (blocking here on this), because some values might be wrapped in double quotes, but some might not be, and this changes from distro to distro for the same value. You need `shlex.split` to parse this:

```python
>>> import pprint, shlex
>>> os_release = {}
>>> with open('/etc/os-release') as file:
...     for line in file:
...         k, v = line.split('=', 1)
...         v, = shlex.split(v)
...         os_release[k] = v
... 
>>> pprint.pprint(os_release)
{'BUG_REPORT_URL': 'https://bugs.debian.org/',
 'HOME_URL': 'https://www.debian.org/',
 'ID': 'debian',
 'NAME': 'Debian GNU/Linux',
 'PRETTY_NAME': 'Debian GNU/Linux 12 (bookworm)',
 'SUPPORT_URL': 'https://www.debian.org/support',
 'VERSION': '12 (bookworm)',
 'VERSION_CODENAME': 'bookworm',
 'VERSION_ID': '12'}
```

You can do it as one-liner, if you don't care about syntactically invalid files:
```python
>>> with open('/etc/os-release') as file:
...     os_release = dict(shlex.split(line)[0].split('=') for line in file)
```
</blockquote></details>

Brilliant one-liner! @jkr0103 Can you replace your regex stuff with the suggested `shlex`? And extract `ID` and `VERSION_ID` from the resulting `os_release` dict.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @jkr0103)

___
*[`gsc.py` line 197 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJFYUT0SW5j9C7Ehey:-NiLvory2nccIMBZBo6i:bn5xhra) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/gsc.py#L197)):*
Ah, and also:

> Note that operating system vendors may choose not to provide version information, for example to accommodate for rolling releases. In this case, `VERSION` and `VERSION_ID` may be unset. Applications should not rely on these fields to be set.

(https://manpages.debian.org/bookworm/systemd/os-release.5.en.html)

Please make sure you handle this case gracefully, i.e. not just traceback a KeyError in the face of the user.

jkr0103 (commented):
I rebased the PR with latest master to take the changes of`redhat/ubi8` support PR which makes the validation job easy.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177#-:-Njq20O_3tT2yGCFRntO:b-z0jjag)* status: 0 of 5 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @woju)

___
*[`gsc.py` line 197 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJFYUT0SW5j9C7Ehey:-Njq2tK37uQSNoO29O9H:b-zbe4ij) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/gsc.py#L197)):*
`os_release` file is identical in both `redhat/ubi8` and `redhat/ubi8-minimal` images hence we had to find other ways to differenciate which can be done by checking the existance of `/usr/bin/microdnf` which exists only for `redhat/ubi8-minimal`.

>So I guess you want to truncate it somewhere, and this depends on distro (alpine would be split at the second dot

Earlier we were compiling Gramine in the docker image base on the distro under `Distro` tag in`config.yaml`. We can compile gramine on the supplied image itself, hence we don't have to retrive exact version of the distro. we only need the distro to point to the right set to template files under `gsc/templates`.

>Please make sure you handle this case gracefully, i.e. not just traceback a KeyError in the face of the user

Done
___
*[`gsc.py` line 184 at r2](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJaSdFBOKP8NgSNa0f:-Njq2XNUEQW1yR0EW4Bh:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/a3e19700c7669d168776b3bede4316dbbcc40719/gsc.py#L184)):*

Brilliant one-liner! @jkr0103 Can you replace your regex stuff with the suggested `shlex`? And extract `ID` and `VERSION_ID` from the resulting `os_release` dict.
</blockquote></details>

Done.

dimakuv (changes requested):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @woju)

___
*[`config.yaml.template` line 13 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvaDyh7GaoEvwuFkX5:-NjvaDyh7GaoEvwuFkX6:bu32b9g) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/config.yaml.template#L13)):*
> ```
> 
> # GSC detects the distro automatically of the supplied image if ditro is set to `auto`.
> # User also have the option to provide one of the supported distros mentioned above.
> ```

In this file, you lost my proposed text rewordings. Please see the difference between revision 2 (r2) and revision 3 (r3) in Reviewable on this file.
___
*[`gsc.py` line 26 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvaQMo7aS5V6uyAWhL:-NjvaQMo7aS5V6uyAWhM:b-g1tvn0) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L26)):*
> ```Python
> 
> # declare constants
> DISTRO_RETRIVAL_ERROR = (f'Could not detect the OS distro of the supplied docker image. Please add '
> ```

`docker` -> `Docker`
___
*[`gsc.py` line 26 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvacBR1d23Pf18ZsS6:-NjvacBR1d23Pf18ZsS7:b-bu4qxb) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L26)):*
> ```Python
> 
> # declare constants
> DISTRO_RETRIVAL_ERROR = (f'Could not detect the OS distro of the supplied docker image. Please add '
> ```

`detect` -> `automatically detect`
___
*[`gsc.py` line 26 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvanGd9B5060v01R1I:-NjvanGd9B5060v01R1J:b3lo446) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L26)):*
> ```Python
> 
> # declare constants
> DISTRO_RETRIVAL_ERROR = (f'Could not detect the OS distro of the supplied docker image. Please add '
> ```

`add` -> `specify`
___
*[`gsc.py` line 27 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvaS_jEEe8OL5np1OB:-NjvaS_jEEe8OL5np1OC:bvnyn4b) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L27)):*
> ```Python
> # declare constants
> DISTRO_RETRIVAL_ERROR = (f'Could not detect the OS distro of the supplied docker image. Please add '
>                          f'OS distro manually in configuration file (Default: config.yaml).')
> ```

Why do you need the `()` around the text? Please remove
___
*[`gsc.py` line 27 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvaXcT5WQIbJn-jlYG:-NjvaXcT5WQIbJn-jlYH:b7nr54b) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L27)):*
> ```Python
> # declare constants
> DISTRO_RETRIVAL_ERROR = (f'Could not detect the OS distro of the supplied docker image. Please add '
>                          f'OS distro manually in configuration file (Default: config.yaml).')
> ```

Why do you need the `f` before each string? You don't use any variables here, so you can just remove the `f` prefix.
___
*[`gsc.py` line 27 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvafgU-8hUkGoF2u8F:-NjvafgU-8hUkGoF2u8G:b-tebn0n) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L27)):*
> ```Python
> # declare constants
> DISTRO_RETRIVAL_ERROR = (f'Could not detect the OS distro of the supplied docker image. Please add '
>                          f'OS distro manually in configuration file (Default: config.yaml).')
> ```

I would remove this `(Default: config.yaml)` completely. Users of GSC should know this info anyway.
___
*[`gsc.py` line 27 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-Njvape03QkkLajf_MTp:-Njvape03QkkLajf_MTq:b-rmj7kp) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L27)):*
> ```Python
> # declare constants
> DISTRO_RETRIVAL_ERROR = (f'Could not detect the OS distro of the supplied docker image. Please add '
>                          f'OS distro manually in configuration file (Default: config.yaml).')
> ```

`in configuration file` -> `in the configuration file`
___
*[`Documentation/index.rst` line 316 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJErFc7VNBRIPZJ0fh:-NjvbsRt1Q_YxW9XhCuv:bef5pfw) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/Documentation/index.rst#L316)):*

Done.
</blockquote></details>

Not done? Looks like you lost this change during rebase.
___
*[`templates/Dockerfile.common.compile.template` line 4 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-Njvc-zjC7LAqLmt8mgD:-Njvc-zjC7LAqLmt8mgE:bve728q) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/templates/Dockerfile.common.compile.template#L4)):*
> ```
> FROM {{Registry}}/{{Distro}} AS gramine
> {% elif app_image %}
> FROM {{app_image}} AS gramine
> ```

Sorry, could you explain what this is?
___
*[`templates/debian/Dockerfile.compile.template` line 41 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-Njvd5wkCPXWKGlhF46C:-Njvd5wkCPXWKGlhF46D:bof5wou) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/templates/debian/Dockerfile.compile.template#L41)):*
> ```
> RUN echo 'deb [arch=amd64] https://download.01.org/intel-sgx/sgx_repo/ubuntu bionic main' \
>     > /etc/apt/sources.list.d/intel-sgx.list \
>     && apt-key add /intel-sgx-deb.key
> ```

Why do you need this change?

jkr0103 (commented):
> Reviewed 1 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: 2 of 5 files reviewed, 12 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[`config.yaml.template` line 13 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvaDyh7GaoEvwuFkX5:-NjwS3Iz-nZ1vklwuvPP:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/config.yaml.template#L13)):*

In this file, you lost my proposed text rewordings. Please see the difference between revision 2 (r2) and revision 3 (r3) in Reviewable on this file.
</blockquote></details>

Done.
___
*[`gsc.py` line 26 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvaQMo7aS5V6uyAWhL:-NjwS58V6wZquYZO5i3f:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L26)):*

`docker` -> `Docker`
</blockquote></details>

Done.
___
*[`gsc.py` line 26 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvacBR1d23Pf18ZsS6:-NjwS5pD2daN4JJsVMO7:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L26)):*

`detect` -> `automatically detect`
</blockquote></details>

Done.
___
*[`gsc.py` line 26 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvanGd9B5060v01R1I:-NjwS7Ei8Pg-0kCULK5a:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L26)):*

`add` -> `specify`
</blockquote></details>

Done.
___
*[`gsc.py` line 27 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvaS_jEEe8OL5np1OB:-NjwS85F3toM7Vf_mxLv:b-6ikowi) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L27)):*

Why do you need the `()` around the text? Please remove
</blockquote></details>

This is required for wraping with 100 characters otherwise it will be one liner.
___
*[`gsc.py` line 27 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvaXcT5WQIbJn-jlYG:-NjwSPMh9UmFIk8cFI92:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L27)):*

Why do you need the `f` before each string? You don't use any variables here, so you can just remove the `f` prefix.
</blockquote></details>

Done.
___
*[`gsc.py` line 27 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-NjvafgU-8hUkGoF2u8F:-NjwSRrL4668toVsAA98:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L27)):*

I would remove this `(Default: config.yaml)` completely. Users of GSC should know this info anyway.
</blockquote></details>

Done.
___
*[`gsc.py` line 27 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-Njvape03QkkLajf_MTp:-NjwSSS26Sm3-h7OLenF:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/gsc.py#L27)):*

`in configuration file` -> `in the configuration file`
</blockquote></details>

Done.
___
*[`Documentation/index.rst` line 316 at r1](https://reviewable.io/reviews/gramineproject/gsc/177#-NiJErFc7VNBRIPZJ0fh:-NjwSfah0spFR4o2tdh_:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/4890c96f8f19776e9ae74e6a67e027570d80dac7/Documentation/index.rst#L316)):*

Not done? Looks like you lost this change during rebase.
</blockquote></details>

Done.
___
*[`templates/Dockerfile.common.compile.template` line 4 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-Njvc-zjC7LAqLmt8mgD:-NjwU9B_9kdJoPKqFVzO:b-qsr3ta) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/templates/Dockerfile.common.compile.template#L4)):*

Sorry, could you explain what this is?
</blockquote></details>

This change will use the supplied application image for Gramine compilation which I feel it correct as it will be the same image where the Gramine build deliverables will be copied and used. No chances of failure of Gramine executable as Gramine was built using exactly the same libraries. If we use some other image for Gramine compilation than that migth fail with some corner cases due to library version mismatch in both the images. e.g. if we Graminize Alpine image `alpine:3.18.4` and uses `alpine:3.18` image for Gramine compilation which have different version of libraries used in Gramine compilation that might cause some issues which execution Gramine deliverables.
___
*[`templates/debian/Dockerfile.compile.template` line 41 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-Njvd5wkCPXWKGlhF46C:-NjwSz5dEOW82xt_FTic:bcs0b6d) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/templates/debian/Dockerfile.compile.template#L41)):*

Why do you need this change?
</blockquote></details>

`intel-sgx-deb.key` is copied at root directory (`/`) few steps above but at this line trying to access it using current working directory. It will fail to access in the images where current working directory is not root directory (`/`). I would say it was a bug.

dimakuv (changes requested):
> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @jkr0103 and @woju)

___
*[`config.yaml.template` line 4 at r4](https://reviewable.io/reviews/gramineproject/gsc/177#-Nk-LfK-A9FefiXWh5Lo:-Nk-LfK-A9FefiXWh5Lp:b-ir6hc5) ([raw file](https://github.com/gramineproject/gsc/blob/c440860bc64350a3361ff8476e6f97ee2009070f/config.yaml.template#L4)):*
> ```
> # gets all tools and dependencies from. This distro should match the distro underlying the
> # application's Docker image; otherwise the results may be unpredictable (if you specify `"auto"`,
> #  which is recommended, you don't need to worry about the mismatch).
> ```

You have one redundant space before `which`
___
*[`gsc.py` line 240 at r4](https://reviewable.io/reviews/gramineproject/gsc/177#-Nk-Pggg-D7EeMs_b9vM:-Nk-Pggg-D7EeMs_b9vN:b-etp4a2) ([raw file](https://github.com/gramineproject/gsc/blob/c440860bc64350a3361ff8476e6f97ee2009070f/gsc.py#L240)):*
> ```Python
>     distro = os_release['ID'] + ':' + os_release['VERSION_ID']
> 
>     if (os_release['ID'] == 'rhel'):
> ```

Can you add a comment before this line:
```
    # RedHat specific logic to distinguish between UBI8 and UBI8-minimal
```
___
*[`gsc.py` line 421 at r4](https://reviewable.io/reviews/gramineproject/gsc/177#-Nk-MZmS3eimogE750EL:-Nk-MZmS3eimogE750EM:bsxzcqa) ([raw file](https://github.com/gramineproject/gsc/blob/c440860bc64350a3361ff8476e6f97ee2009070f/gsc.py#L421)):*
> ```Python
>     distro = env.globals['Distro']
>     if distro == 'auto':
>         print('Please specify the Distro in the configuration file.')
> ```

This is cryptic, I suggest to use the same text as in the check for `Image` above:
```
        print('`gsc build-gramine` does not allow `Distro` set to `auto`.')
```
___
*[`templates/Dockerfile.common.compile.template` line 4 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-Njvc-zjC7LAqLmt8mgD:-Nk-Oo7W-umyHrCViPly:b-aqnc2g) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/templates/Dockerfile.common.compile.template#L4)):*

This change will use the supplied application image for Gramine compilation which I feel it correct as it will be the same image where the Gramine build deliverables will be copied and used. No chances of failure of Gramine executable as Gramine was built using exactly the same libraries. If we use some other image for Gramine compilation than that migth fail with some corner cases due to library version mismatch in both the images. e.g. if we Graminize Alpine image `alpine:3.18.4` and uses `alpine:3.18` image for Gramine compilation which have different version of libraries used in Gramine compilation that might cause some issues which execution Gramine deliverables.
</blockquote></details>

Sorry, but this change is **very** controversial.

Please split this change into a separate PR, and let's discuss it in that separate PR.

In general, I don't like this change because there are many issues it leads to:
- Why do we need the `Distro` parameter at all then? We always have the `app_image` supplied in `gsc build`, so what's even the point of `Distro` then?
- What if the `app_image` has a really weird original image, such that GSC can't detect what it is? Then we're doomed to fail always, based on your current logic. However, in the previous GSC logic, the user can play with `Distro` and find something working.
- This change seems to be done purely to support Alpine versioning? I agree that it is a problem, but the current solution breaks other things...
- Why if we specify `Registry`, then we never hit this case? It's illogical.
___
*[`templates/debian/Dockerfile.compile.template` line 41 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-Njvd5wkCPXWKGlhF46C:-Nk-OO-8-UVVGvd7g2f6:b-rv35d9) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/templates/debian/Dockerfile.compile.template#L41)):*

`intel-sgx-deb.key` is copied at root directory (`/`) few steps above but at this line trying to access it using current working directory. It will fail to access in the images where current working directory is not root directory (`/`). I would say it was a bug.
</blockquote></details>

This is a good one! Indeed, it's a bug.

But can we please separate it into another PR? And we'll merge that PR asap. Please do it, with the explanation (in the PR description and in the commit message) as you described here.

We definitely don't want to stash unrelated changes to one PR.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: 1 of 5 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[`config.yaml.template` line 4 at r4](https://reviewable.io/reviews/gramineproject/gsc/177#-Nk-LfK-A9FefiXWh5Lo:-Nk0TXdc-gjUmtfdWFWQ:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/c440860bc64350a3361ff8476e6f97ee2009070f/config.yaml.template#L4)):*

You have one redundant space before `which`
</blockquote></details>

Done.
___
*[`gsc.py` line 240 at r4](https://reviewable.io/reviews/gramineproject/gsc/177#-Nk-Pggg-D7EeMs_b9vM:-Nk0TeHdALbW9LPb1PCi:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/c440860bc64350a3361ff8476e6f97ee2009070f/gsc.py#L240)):*

Can you add a comment before this line:
```
    # RedHat specific logic to distinguish between UBI8 and UBI8-minimal
```
</blockquote></details>

Done.
___
*[`gsc.py` line 421 at r4](https://reviewable.io/reviews/gramineproject/gsc/177#-Nk-MZmS3eimogE750EL:-Nk0TiRx-QDFbw3vsHkQ:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/c440860bc64350a3361ff8476e6f97ee2009070f/gsc.py#L421)):*

This is cryptic, I suggest to use the same text as in the check for `Image` above:
```
        print('`gsc build-gramine` does not allow `Distro` set to `auto`.')
```
</blockquote></details>

Done.
___
*[`templates/Dockerfile.common.compile.template` line 4 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-Njvc-zjC7LAqLmt8mgD:-Nk0Tv7y3M5MvB_dcpO4:bdqwq2c) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/templates/Dockerfile.common.compile.template#L4)):*

Sorry, but this change is **very** controversial.

Please split this change into a separate PR, and let's discuss it in that separate PR.

In general, I don't like this change because there are many issues it leads to:
- Why do we need the `Distro` parameter at all then? We always have the `app_image` supplied in `gsc build`, so what's even the point of `Distro` then?
- What if the `app_image` has a really weird original image, such that GSC can't detect what it is? Then we're doomed to fail always, based on your current logic. However, in the previous GSC logic, the user can play with `Distro` and find something working.
- This change seems to be done purely to support Alpine versioning? I agree that it is a problem, but the current solution breaks other things...
- Why if we specify `Registry`, then we never hit this case? It's illogical.
</blockquote></details>

Sure, will create separate PR for this.
___
*[`templates/debian/Dockerfile.compile.template` line 41 at r3](https://reviewable.io/reviews/gramineproject/gsc/177#-Njvd5wkCPXWKGlhF46C:-Nk0U4DGCeDrtfwf96q4:b2i62v7) ([raw file](https://github.com/gramineproject/gsc/blob/af9b53897e220be021a533da032d2b75bce645c3/templates/debian/Dockerfile.compile.template#L41)):*

This is a good one! Indeed, it's a bug.

But can we please separate it into another PR? And we'll merge that PR asap. Please do it, with the explanation (in the PR description and in the commit message) as you described here.

We definitely don't want to stash unrelated changes to one PR.
</blockquote></details>

Moved the change in separate PR https://github.com/gramineproject/gsc/pull/179

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[`gsc.py` line 240 at r5](https://reviewable.io/reviews/gramineproject/gsc/177#-NkEap_e9jfYFou9OZKy:-NkK_FvYEXbYZkALa5mp:b-ffhhmh) ([raw file](https://github.com/gramineproject/gsc/blob/e5d9f2f21f2db6c30f6c4373861d80e9b3921c0f/gsc.py#L240)):*

I like this change (and don't block), but we don't have Alpine? Is there some plan to add it to GSC?
</blockquote></details>

We have it in the backlogs and might take in the next release.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @woju)

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: 4 of 5 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @woju)

___
*[`gsc.py` line 26 at r5](https://reviewable.io/reviews/gramineproject/gsc/177#-NkKgxEsARYTviLqjVuF:-NkOyhuj7cBidjvru8Ry:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/e5d9f2f21f2db6c30f6c4373861d80e9b3921c0f/gsc.py#L26)):*

Please make this an `Exception`:

```python
class DistroRetrievalError(Exception):
    def __init__(self, *args):
        super().__init__('that long error message', *args)

def foo():
    raise DistroRetrievalError()

def bar():
    try:
        foo()
    except DistroRetrievalError as e:
        print(e, file=sys.stderr)
```
</blockquote></details>

Done.
___
*[`gsc.py` line 249 at r5](https://reviewable.io/reviews/gramineproject/gsc/177#-NkKy1eO0HCwVl5YKTMu:-NkOyn_i5P4UdMXQHjbg:b-896fix) ([raw file](https://github.com/gramineproject/gsc/blob/e5d9f2f21f2db6c30f6c4373861d80e9b3921c0f/gsc.py#L249)):*
<details><summary><i>Previously, kailun-qin (Kailun Qin) wrote…</i></summary><blockquote>

Should we catch the specific exception, e.g., `docker.errors.ContainerError` here?
</blockquote></details>

Done.
___
*[`gsc.py` line 492 at r5](https://reviewable.io/reviews/gramineproject/gsc/177#-NkKj21E0sN4j806mdbJ:-NkOypKg1efCWiXFfKAx:bdlmvpp) ([raw file](https://github.com/gramineproject/gsc/blob/e5d9f2f21f2db6c30f6c4373861d80e9b3921c0f/gsc.py#L492)):*

Yeah, definitely make that `DistroRetrievalError`. Catching `KeyError` on such a general level is very wrong, it's a footgun for other programmers, because it makes debugging harder, e.g. if you misspelled the key name. `LookupError`s (`KeyError`s and `IndexError`s) should be catched in the same function where the access happens, with very few exceptions, and if you need some exception to propagate, make yourself a custom exception.
</blockquote></details>

We are just printing the error message and exiting, hence catching all exception under one.

dimakuv (approved):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @woju)

kailun-qin (approved):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @woju)

dimakuv (approved):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/177)* status: :shipit: complete! all files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/177#-Nl2b8EGBFvbXUhXyz1L:-Nl2b8EGBFvbXUhXyz1M:ba4d0og) (no related file):*
I tested this PR (the final rebased version) against different Python distros:
- `python` (based on Bookworm) works
- `python:bullseye` works
- `python:alpine` correctly prints an error: "`alpine` distro is not supported by GSC."

Looks good to me.

___

PR #208, opened 2024-07-15
Add multiprocessing lock in prune Docker images

adarshan-intel:
## Description of the changes 

This PR adds a lock to the Docker image pruning process to prevent race conditions and breaking the code

## How to test this PR? 

To test this PR, create a script that signs multiple images in parallel in background using `&`. Then, attempt to trigger the race condition while performing the `gsc sign` operation. This should help you encounter the issue.
Example script [Link](https://gist.github.com/adarshan-intel/97a1e7d1bce956f8fb7015438d9bda24)

image

Adding locking mechanism should fix the issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gsc/208)

### DIFF
--- a/gsc.py
+++ b/gsc.py
@@ -6,6 +6,7 @@
 
 import argparse
 import hashlib
+import multiprocessing
 import os
 import pathlib
 import re
@@ -521,7 +522,9 @@ def gsc_sign_image(args):
         os.remove(tmp_build_key_path)
         # Remove a temporary image created during multistage docker build to save disk space.
         # Please note that removing the image doesn't assure security.
-        docker_socket.api.prune_images(filters={'label': 'build_id=' + build_id})
+        lock = multiprocessing.Lock()
+        with lock:
+            docker_socket.api.prune_images(filters={'label': 'build_id=' + build_id})
 
     if get_docker_image(docker_socket, signed_image_name) is None:
         print(f'Failed to build a signed graminized Docker image `{signed_image_name}`.')


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/208)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @adarshan-intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/208#-O1pbh2G4GhkqJ6dVmio:-O1pbh2G4GhkqJ6dVmip:b-25oxpe) (no related file):*
Why fixing this?

First of all, this seems to be a problem of Docker itself, which is acknowledged and planned to be fixed (IIUC): https://github.com/moby/moby/issues/36447

Second, what is the scenario that you try to fix/circumvent? Why would anyone want to sign multiple images in parallel?

___

mkow:
Third, please never paste text as screenshots of it...

aneessahib:
@dimakuv  -  A customer has reported an issue where their build system is designed to execute multiple jobs concurrently to generate several independent GSC images. Although this approach will result in a sequential process, I believe it serves as an effective temporary solution to resolve their current situation.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gsc/208)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @adarshan-intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gsc/208#-O1pbh2G4GhkqJ6dVmio:-O1uHd7mBiZM_GP7xv6f:b-8kg4l3) (no related file):*
> A customer has reported an issue where their build system is designed to execute multiple jobs concurrently to generate several independent GSC images. Although this approach will result in a sequential process, I believe it serves as an effective temporary solution to resolve their current situation.

Thanks Anees for the explanation.

However, I don't think this can be fixed at the GSC level. The current solution (in this PR) can't work (unless I misunderstand how Python locks work). The fix should happen at the customer's build system level, I think.

___
___
*[`gsc.py` line 526 at r1](https://reviewable.io/reviews/gramineproject/gsc/208#-O1uHyek7Th22rf6Nidq:-O1uHyek7Th22rf6Nidr:b-r0se9v) ([raw file](https://github.com/gramineproject/gsc/blob/127e463877b7053990bc5bb98880756dcf67f72b/gsc.py#L526)):*
> ```Python
>         # Please note that removing the image doesn't assure security.
>         lock = multiprocessing.Lock()
>         with lock:
> ```

How can this work? You're effectively trying to synchronize independent GSC processes. These independent processes do not share any state, and do not constitute a group of processes, so their locking operations cannot affect each other.

I'm 99% sure this change doesn't have an effect on the data race the customer observes.

If I'm mistaken, please explain to me how `multiprocessing.lock` works then; in particular, how this locking affects **all** other Python processes.

